PR #138, opened 2015-01-09
Don't set OPENSSL_VERSION to _OPENSSL_VERSION

eagleflo:
[Casablanca](https://casablanca.codeplex.com/) is a project that uses CMake as it's build tool. On OS X, it tries to prefer the OpenSSL version found from Homebrew over the system-supplied one with the following CMake snippet:

```
  if(APPLE AND NOT OPENSSL_ROOT_DIR)
    # Prefer a homebrew version of OpenSSL over the one in /usr/lib
    file(GLOB OPENSSL_ROOT_DIR /usr/local/Cellar/openssl/*)
    # Prefer the latest (make the latest one first)
    list(REVERSE OPENSSL_ROOT_DIR)
  endif()
  # This should prevent linking against the system provided 0.9.8y
  find_package(OpenSSL 1.0.0 REQUIRED)
```

However, this fails due to FindOpenSSL module blindly setting OPENSSL_VERSION to _OPENSSL_VERSION, even when OPENSSL_INCLUDE_DIR at that point already points to the Homebrew-supplied version.

(Workaround for previous versions of CMake is to set _OPENSSL_VERSION to "" just before `find_package`.)

### DIFF
--- a/Modules/FindOpenSSL.cmake
+++ b/Modules/FindOpenSSL.cmake
@@ -283,9 +283,7 @@ function(from_hex HEX DEC)
 endfunction()
 
 if (OPENSSL_INCLUDE_DIR)
-  if (_OPENSSL_VERSION)
-    set(OPENSSL_VERSION "${_OPENSSL_VERSION}")
-  elseif(OPENSSL_INCLUDE_DIR AND EXISTS "${OPENSSL_INCLUDE_DIR}/openssl/opensslv.h")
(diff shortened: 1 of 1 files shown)


eagleflo:
Might be related to http://www.cmake.org/Bug/bug_relationship_graph.php?bug_id=15075.

bradking:
Thanks.  Applied as commit 99b13820715ff3b98c2c0cb114b38620b6da4466.

PR #141, opened 2015-02-02
Xcode: Allow user decide where to put build.

thedmd:
Allow user to put build directory in different location than Xcode default. If in CMake files SYMROOT is set by user use it instead of reverting user choice.


bradking:
The `CMAKE_XCODE_ATTRIBUTE_` feature is a hack to help users set things that CMake does not know how to generate.  If there is a use case for changing the SYMROOT, it should be a first-class feature of CMake to tell it to do so.  Other things in CMake may depend on knowing where the binaries are going.

Please read CONTRIBUTING.rst and bring this up for discussion on the mailing list.

PR #142, opened 2015-02-03
Fixed bad include in FindProtobuf.cmake

impala454:
Line 243 which includes FindPackageHandleStandardArgs fails in Windows under CMake 3.1.0.  Removing the directory prefix and file extension fixed it.  Confirmed working in Windows and Ubuntu 14.04.1.


bradking:
All find modules include FPHSA that way.  What is the actual error you see?

impala454:
Sure but by specifying the list directory isn't necessary.  This was a specific case where I found the problem when specifying a different CMAKE_MODULE_PATH.  FindProtobuf.cmake was not included with my CMake so I pulled it into my project.  This change makes the module work anywhere instead of just in the default directory.  Perhaps it's worth changing in others as well.

impala454:
The specific error is that it can't find the file C:/path/to/project/cmake/module/dir/FindPackageHandleStandardArgs.cmake

bradking:
So it works within CMake but not when copying the module out.  This is expected.  If you copy the module out you can edit this line, but the copy within CMake wants to make sure it gets the FPHSA it expects.  In the past we did not do this and ended up having to create [CMP0017](http://www.cmake.org/cmake/help/v3.1/policy/CMP0017.html) to protect against out-of-date copies of CMake builtin modules in the project CMAKE_MODULE_PATH.

Note when copying a module there the copyright notice has instructions below it about what to do with the license.

impala454:
The module was not included to begin with in CMake for Windows 3.1.0.  Is there some way I can get this the "proper" way?

bradking:
Huh? FindProtobuf has been distributed with CMake since 2.8.0.  I see it in my 3.1 installation.

impala454:
Sheesh.  I do not know what happened here, I do see it now.  I have no idea why I was having the issue but I can't seem to recreate it.  My apologies.

PR #144, opened 2015-02-19
Update FindJsonCpp.cmake

rpavlik:
JsonCpp can generate a CMake config file during its build. This change adapts the find module to check for that config file first. I did try to provide compatibility so that build systems don't have to care whether they're getting the data from normal find processes or a config file: setting up the INCLUDE and LIBRARY variables (though not cache variables), which should correspondingly permit the version detection to proceed as well.

I'm not sure what the best practice is for effectively providing "aliases" of imported targets, but I gave it my best shot here: jsoncpp's built-in config file defines an imported target `jsoncpp_lib`. Of course, the config file knows if it's a static or shared library, so my definition of the new imported target as `UNKNOWN` is probably sub-optimal. Initially tried to make it an interface library, but it seems that the "namespaced" names are not permitted in anything except imported libraries.

Also not sure how best to handle the `NOTFOUND` cache variables if you configure once, with it not finding the library, then you set `CMAKE_PREFIX_PATH` or `jsoncpp_DIR` accordingly so it finds the config file: it leaves those sitting in the cache right now, potentially confusingly.

Also, at the moment it will always using the config file even if those cache vars are manually set - not sure if this is OK behavior.

Not sure how much of this tweaking should be resolved before incorporating this patch, but when I saw mention of the jsoncpp module in pre-release notes, I jumped to try to get this in ASAP - and I'm both more comfortable with GitHub pull request/issues than Mantis and I expect actual patches are preferred to bare bug reports.

### DIFF
--- a/Modules/FindJsonCpp.cmake
+++ b/Modules/FindJsonCpp.cmake
@@ -2,7 +2,9 @@
 FindJsonCpp
 -----------
 
-Find JsonCpp includes and library.
+Find JsonCpp includes and library. If JsonCpp was built with the
+``JSONCPP_WITH_CMAKE_PACKAGE`` option enabled, generating a config file,
+it is used preferentially.
 
 Imported Targets
 ^^^^^^^^^^^^^^^^
@@ -42,6 +44,8 @@ This module uses the following cache variables:
   The location of the JsonCpp include directory containing ``json/json.h``.
 
 The cache variables should not be used by project code.
+They may not be set (or may be set to ``...-NOTFOUND``) if the config
+file method was used successfully to find the library.
 They may be set by end users to point at JsonCpp components.
 #]=======================================================================]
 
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  From a quick glance the `jsoncppConfig.cmake` file that upstream JsonCpp produces is not quite sufficient to provide all the information the FindJsonCpp module needs.  We cannot rely on specific properties like `INTERFACE_INCLUDE_DIRECTORIES` being set because CMake makes no guarntees about how the interfaces are implemented under the hood.  Future versions may export the targets differently.  Instead they should export the targets into a Targets.cmake file and load that from a separately generated Config.cmake file that also sets information like include directories in well-defined, documented variables.

Still, it would be nice to have the FindJsonCpp module use jsoncpp_DIR if it can.  For broader discussion, please read `CONTRIBUTING.rst` and bring this patch to the mailing list.

rpavlik:
Thanks for your quick response and feedback! I've been on the CMake list a long time (and was pretty active for a while), just haven't been keeping up with it or much of my personal email recently (graduation, new job, that sort of thing). Sorry for not following "protocol" on this one, I promise I'm usually a better project contributor than this :smile: 

I'll be pointing the jsoncpp folks to this pull request in hopes of getting their generated files to meet the requirements. As it stands, AFAICT, on case-insensitive filesystems, unless I specifically change `find_package(jsoncpp)` as I have now to `find_package(jsoncpp CONFIG)` in my project build files, my builds will break when used with the 3.2.0 release.

bradking:
Since having upstreams provide package config files is preferable to maintaining a find module within CMake, I've created commit a5768442636c7fe909e8afc205fd19ac13b9fbc2 to revert providing FindJsonCpp publicly in the 3.2 release.  It had been added only because we needed it for building CMake itself with new features implemented with jsoncpp.

Instead we should focus on getting jsoncpp to provide good cmake package files unconditionally.

rpavlik:
Thanks - this sounds like a good solution.

PR #147, opened 2015-03-19
PROTOBUF_GENERATE_CPP: fix generated paths

mantiz:
The "problem" here is when you place your `*.proto` files inside a subdirectory, e.g. `proto`.
The call to `protoc` will create the subdirectory `proto` inside of `CMAKE_CURRENT_BINARY_DIR`, so the resulting path is something like `${CMAKE_CURRENT_BINARY_DIR}/proto/example.pb.*` but the function will add `${CMAKE_CURRENT_BINARY_DIR}/example.pb.*` to the lists.

So, if you have multiple `*.proto` files with the same name in different directories, this might be a problem because the same path would be added to the list.

I am not 100% sure if this is really a problem or if it would simply result in regenerating the source files with each build even if it is not necessary.

### DIFF
--- a/Modules/FindProtobuf.cmake
+++ b/Modules/FindProtobuf.cmake
@@ -127,13 +127,14 @@ function(PROTOBUF_GENERATE_CPP SRCS HDRS)
   foreach(FIL ${ARGN})
     get_filename_component(ABS_FIL ${FIL} ABSOLUTE)
     get_filename_component(FIL_WE ${FIL} NAME_WE)
+    get_filename_component(DIR ${FIL} DIRECTORY)
 
-    list(APPEND ${SRCS} "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.pb.cc")
-    list(APPEND ${HDRS} "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.pb.h")
+    list(APPEND ${SRCS} "${CMAKE_CURRENT_BINARY_DIR}/${DIR}/${FIL_WE}.pb.cc")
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  I agree this is how the module should have behaved originally.  Unfortunately projects may have come to depend on the current behavior.  Moving the generated source file may not be a problem.   However, project source code will have header file search paths (`-I`) and `#include` lines written assuming the current location of the headers.  Some kind of option will be needed to enable this behavior.

mantiz:
Hm, didn't think about that.

I simply add `${CMAKE_CURRENT_BINARY_DIR}` to my search paths and ignore the variables filled by this function. :)

However, I created another small test case and found some more problems related to this.
A call like this
```
protobuf_generate_cpp(SOURCES HEADERS test1/test.proto test2/test.proto)
```
results in
```
/usr/bin/protoc --cpp_out /home/mantiz/tmp/proto-test/build -I /home/mantiz/tmp/proto-test/test1 -I /home/mantiz/tmp/proto-test/test2 /home/mantiz/tmp/proto-test/test2/test.proto
```
which fails with the following error message:
```
Input is shadowed in the --proto_path by "/home/mantiz/tmp/proto-test/test1/test.proto".  Either use the latter file as your input or reorder the --proto_path so that the former file's location comes first.
```
So, the "fix" I proposed is incomplete.
I can dig a bit deeper and try to solve this if you are interested in this change.

It's ok for me to just close this PR for now since I think there is no problem in most cases and I think the complete fix will involve bigger changes and would therefore better be proposed in another PR.

What do you think?

bradking:
> complete fix will involve bigger changes and would therefore better be proposed in another PR

I agree.  Please look into the more complete fixes.  Also, please read `CONTRIBUTING.rst`.

PR #150, opened 2015-04-01
Set Xcode file type for Metal shader files

seanjames777:
Sets the Xcode file type to "sourcecode.metal". Xcode can then compile and embed Metal shaders into applications. Note that Metal shaders need to be listed as Resources, i.e.:

set_target_properties(test_ios PROPERTIES RESOURCE ../content/shaders/shaders.metal)


bradking:
Thanks.  Applied with minor tweak as commit 078c732185ff610c55707ae66be0a787c78bd2e2.

PR #151, opened 2015-05-05
QMacInstallDialog: Use Authorization Services to install command line to...

oyarzun:
Use authorization services to install commad line tool symlinks on OSX.
related to bug http://public.kitware.com/Bug/print_bug_page.php?bug_id=10056
http://stackoverflow.com/questions/23849962/cmake-installer-for-mac-fails-to-create-usr-bin-symlinks

### DIFF
--- a/Source/QtDialog/QMacInstallDialog.cxx
+++ b/Source/QtDialog/QMacInstallDialog.cxx
@@ -34,94 +34,43 @@ QMacInstallDialog::~QMacInstallDialog()
 
 void QMacInstallDialog::DoInstall()
 {
-  OSStatus myStatus;
-
-	AuthorizationFlags myFlags = kAuthorizationFlagDefaults;
-  AuthorizationRef myAuthorizationRef;
-  myStatus = AuthorizationCreate(NULL, kAuthorizationEmptyEnvironment,
-                                 myFlags, &myAuthorizationRef);
-  if (myStatus != errAuthorizationSuccess) return;
-  
-  AuthorizationItem myItems = {kAuthorizationRightExecute, 0,
-	                                 NULL, 0};
-	AuthorizationRights myRights = {1, &myItems};
-	myFlags = kAuthorizationFlagDefaults |
-	          kAuthorizationFlagInteractionAllowed |
-	          kAuthorizationFlagPreAuthorize |
-	          kAuthorizationFlagExtendRights;
-	myStatus =
-	      AuthorizationCopyRights(myAuthorizationRef,&myRights,NULL,myFlags,NULL);
-	  
-	if (myStatus == errAuthorizationSuccess)
-  {
-    QDir installDir(this->Internals->InstallPrefix->text());
-    QString installTo = installDir.path();
-    if(!cmSystemTools::FileExists(installTo.toLocal8Bit().data()))
+  std::string cmd;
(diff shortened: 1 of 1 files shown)


clintonstimpson:
Thanks for looking into this issue.
One concern I have with this patch is the use of the deprecated AuthorizationExecuteWithPrivileges() which has been deprecated for quite some time.

I understand that the alternative is to have a helper application, which then required code signing.
However, have you looked into using a shell script to solve this problem?  /bin/bash should already be code signed and it can simply run the script or commands.
Will that approach work, and also use non-deprecated APIs?

bradking:
Rather than trying to implement this at all in the GUI, perhaps the menu option should instead just pop up instructions on how to do it via the command line.  I expect anyone wanting to install the command line interface is willing to use the command line with `sudo /Applications/CMake.app/Contents/bin/cmake-gui --install-command-line` or something like that.

FWIW, I just put `/Applications/CMake.app/Contents/bin` in my `PATH`.

oyarzun:
I've updated the pull request to use apple script to install the symlinks as root without using a deprecated api.

clintonstimpson:
On May 6, 2015 12:16 PM, oyarzun <notifications@github.com> wrote:
>
> I've updated the pull request to use apple script to install the symlinks as root without using a deprecated api.
>
That does look better.  Thanks.
Clint

bradking:
Rather than using `system` to run the command, please use `cmSystemTools::RunSingleCommand` (the overload that takes `vector<string>` as the command line).  That allows one to capture the output and check the result of the operation.  Use this to restore the message box on failure.

clintonstimpson:
I too want to see Brad's suggested changes.
One question... if one is already root/administrator, will it still prompt for administrative rights?

bradking:
> if one is already root/administrator, will it still prompt

Furthermore, if one chooses a destination directory that does not require admin rights, we shouldn't prompt for them.

My above suggestion of requiring a manual `sudo` command line avoids all such complexity.

oyarzun:
No it will not prompt you again if you are running the GUI as root. 

From what I have seen, OSX GUI applications which install command line tools either provide you with a menu option to install the tools or have an pkg installer. The menu will install the tools to either /usr/bin or /usr/local/bin without providing the option to pick a directory. Atom and Unison are ones that use a menu item.

Should I update the pull request to use cmSystemTools::RunSingleCommand? Or just change it to provide the instructions to run the gui with sudo (which to your point would only be required if they wanted it in a dir that required admin rights)?

bradking:
My preference is to remove the current menu option and add a different one in an appropriate place to get help for installing the symlinks.  Then provide instructions for command-line installation, e.g.

    sudo /Applications/CMake.app/Contents/bin/cmake-gui --install-command-line[=/custom/dir]

Of course the path in the command line would be computed from the actual location of the executable.

This approach is similar to `xcode-select --install` for Xcode command-line tools.

clintonstimpson:
By the way, there is also a way to install Xcode command-line tools from within the Xcode GUI (Xcode Preferences -> Downloads).

However, I like Brad's suggestion over the current patch.

bradking:
I've implemented the command-line approach in commit da9dddb09020e1dbeb6ec4d29bf9d97ec796102e and commit 012357cb2cf255aec304905901cf6806849d94b3.

PR #152, opened 2015-05-07
Add modules for SDL2, SDL2_image, SDL2_mixer, SDL2_net, SDL2_ttf.

Quent42340:
The SDL2 is currently used a lot, and it's too bad that CMake doesn't include a portable way to find them.

### DIFF
--- a/Modules/FindSDL2.cmake
+++ b/Modules/FindSDL2.cmake
@@ -1,0 +1,163 @@
+# Locate SDL2 library
+# This module defines
+# SDL2_LIBRARY, the name of the library to link against
+# SDL2_FOUND, if false, do not try to link to SDL2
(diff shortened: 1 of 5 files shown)


bradking:
Thanks for working on this.  However, as discussed here:

  http://www.cmake.org/Bug/view.php?id=14826#c35508

since SDL2 builds with CMake they should provide package configuration files as documented in the `cmake-packages(7)` manual:

  http://www.cmake.org/cmake/help/v3.2/manual/cmake-packages.7.html

CMake upstream should not have to maintain builtin knowledge about every library of every package out there, especially not those that are already CMake-aware.  The place to contribute support for using SDL2 from CMake-based applications is to SDL2.

PR #154, opened 2015-05-08
Added FindSFML.cmake

TankOs:
Request for adding SFML library search support to CMake.

### DIFF
--- a/Modules/FindSFML.cmake
+++ b/Modules/FindSFML.cmake
@@ -1,0 +1,366 @@
+# This script locates the SFML library
+# ------------------------------------
+#
+# Usage
+# -----
+#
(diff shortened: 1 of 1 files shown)


bradking:
Thanks for working on this.  However, since SFML builds with CMake they should provide package configuration files as documented in the `cmake-packages(7)` manual:

  http://www.cmake.org/cmake/help/v3.2/manual/cmake-packages.7.html

CMake upstream should not have to maintain builtin knowledge about every library of every package out there, especially not those that are already CMake-aware.  The place to contribute support for using SFML from CMake-based applications is to SFML.

TankOs:
Thank you for the info, @bradking

PR #155, opened 2015-05-17
Fix findpostgresql

takar:
Fixes the FindPostgreSQL module on recent Debian and Ubuntu systems

### DIFF
--- a/Modules/FindPostgreSQL.cmake
+++ b/Modules/FindPostgreSQL.cmake
@@ -170,7 +170,7 @@ if (PostgreSQL_INCLUDE_DIR)
 endif()
 
 # Did we find anything?
-include(FindPackageHandleStandardArgs)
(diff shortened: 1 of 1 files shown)


bradking:
Thanks for working on this.  I'll comment on individual commits.

bradking:
Thanks for reverting the relocatable part.  Please also see my comment in takar/CMake@75af1aa17469b2035756f8d73a9b95d762f1aafe about use of `PATH_SUFFIXES` instead of hard-coded paths.

takar:
Hi Brad, thanks for your comments. I agree that the hard-coded paths should be avoided. I'll propose an improved patch later this week.

bradking:
Closing for now.  Please comment when you have updates or read `CONTRIBUTING.rst` and send a patch tothe mailing list.  Thanks!

PR #157, opened 2015-05-28
Add DEFINES feature for FindBISON

Uni-:
This changeset adds `DEFINES` parameter feature to FindBISON module (Modules/FindBISON.cmake).

The `BISON_TARGET` macro spec changes from:
```plain
BISON_TARGET(<Name> <YaccInput> <CodeOutput> [VERBOSE <file>]
            [COMPILE_FLAGS <string>])
```
to:
```plain
BISON_TARGET(<Name> <YaccInput> <CodeOutput> [VERBOSE <file>]
            [COMPILE_FLAGS <string>] [DEFINES <file>])
```

The `DEFINES <file>` parameter causes the macro-defined bison command to contain `--defines=<file>` rather than `-d`, so the header file name is specified.

While `--defines` option can be given via `COMPILE_FLAGS`, doing it messes the whole build process by running the `BISON_TARGET` macro again and again. It's becuase the header file name `BISON_${Name}_OUTPUT_HEADER` is generated independently, so the command recognizes wrong output `BISON_${Name}_OUTPUT`.

In this changeset, `DEFINES <file>` parameter not only invokes `--defines`, but also changes `BISON_${Name}_OUTPUT_HEADER` and `BISON_${Name}_OUTPUT`.

Usage:
```plain
BISON_TARGET(MyParser myparser.y myparser.tab.cc DEFINES myparser.tab.h)
```


bradking:
Thanks for working on this.  I think the `DEFINES` parameter would be better named `DEFINES_FILE` to clarify the reason a file name follows it.

The `# Parsing parameters` block in the existing code is not our modern way to parse macro arguments.  Rather than extending it to parse the new parameter, please look at converting the parameter parsing over to use the `CMakeParseArguments` module or use a manual state machine loop (see `FortranCInterface_HEADER` for an example).

Please also read `CONTRIBUTING.rst`.

Uni-:
Hi @bradking, all of your suggestions are done in the second commit; also I sent the patches via the cmake-developers mailing list.

bradking:
Thanks.  For reference, another link to the mailing list thread is [DEFINES_FILE feature for FindBISON module](http://thread.gmane.org/gmane.comp.programming.tools.cmake.devel/13335).  Further review will take place there so I'm closing this PR.

PR #159, opened 2015-06-15
Fix `GNUInstallDirs.cmake` according GNU Coding Standard and Filesystem Hierarchy Standard

zaufi:
I've faced with this bug a long time ago, but in my projects I always do the workarounds. Unfortunately I've slapped by this bug recently from Debian schroot package, which is recently started to use CMake (and seem developers have no enough experience w/ it). So, I've done this patch to help further CMake users to produce packages that behave in the expected way...


Tests/CMakeOnly/GNUInstallDirs2Other/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
+cmake_minimum_required(VERSION 3.2.3)
+
+project(GNUInstallDirsTest)
```
bradking: These tests do not actually compile any code so they should be able to use the `NONE` option to the `project()` command to avoid enabling languages.  This will make them much faster, especially on IDE generators like Visual Studio.

Tests/CMakeOnly/GNUInstallDirs2Other/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+    DATADIR
+    INFODIR
+    LOCALEDIR
+    MANDIR
+    DOCDIR
+  )
+
+set(
+    SHORT_DIR_LIST
+    bin
```
bradking: The expected values could be placed in a table like

    set(expect_short_BINDIR "bin")
    set(expect_long_BINDIR "/usr/local/bin")

and then referenced below as `${expect_short_${dir}}` and `${expect_long_${dir}}`.

bradking:
Thanks for working on this.  Please also read `CONTRIBUTING.rst`.

Tests/CMakeOnly/GNUInstallDirs2Other/CMakeLists.txt
zaufi: Unfortunately it'll cause an error like this:

    CMake Warning (dev) at /usr/share/cmake/Modules/GNUInstallDirs.cmake:177 (message):
      Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
      target architecture is known.  Please enable at least one language before
      including GNUInstallDirs.

zaufi: Ok, there is a workaround: predefine `CMAKE_SIZEOF_VOID_P` -- it'll be ignored anyway, cuz otherwise (asserting `LIBDIR`) would make tests "unpredictable".

bradking:
I've just revised the existing module documentation in commit 1199ebf1c53f7f591e412617afdaaf4ccd5bc0ff.  Please rebase on that commit.  I left the word "typically" in an appropriate place so that you can add a following sentence that explains there are special cases documented below.

Tests/CMakeOnly/GNUInstallDirs2Other/CMakeLists.txt
zaufi: Fixed

Tests/CMakeOnly/GNUInstallDirs2Other/CMakeLists.txt
zaufi: Fixed

zaufi:
It seems I've done smth wrong... I've got no idea what to do now, after I pulled changes from upstream's master into my fork. Now PR show that I want to pull a lot of changes...
I'm giving up...

zaufi:
Yep, I've read `CONTRIBUTING.rst` before submitting my PR -- it is why I've done this patch in a `master`. But now I have no idea how to merge it w/ your revised docs commit...

bradking:
Okay, I've taken your changes and factored them out into my own sequence of commits.  After further testing I will merge the changes and link from here.

bradking:
See commits 5f30f1754ac9a701cbd311bab86250dd237d86fd and c8bd37ec685c02736618af83ac894e96fc1e6ab8.  Thanks for this contribution!

zaufi:
Glad to hear that. Looking forward for next release to remove workarounds from my projects…

PR #146, opened 2015-03-06
Emacs mode: Several additions & adjustments

roysc:
Define cmake-mode properly with `define-derived-mode`
Add font-lock for keywords and variables
Rename to `cmake-unscreamify-buffer`
`cmake-indent`: Make closing parens match opening; clean up some

PR #160, opened 2015-06-23
Make cmake-mode support prog-mode hooks.

josteink:
Since version24, Emacs supports a generic mode called `prog-mode`. Like all other modes it has its own mode-hook, `prog-mode-hook`. For emacs users it's common to provide all your generic programming-mode related customizations in this mode-hook.

`cmake-mode` is definitely a programming-mode and should support calling this hook. There are two ways to make that happen:

* Make your major-mode a derived-mode from `prog-mode`.
* Manually calling the hook upon mode-activation.

Implementing a derived mode may be the most proper thing to do (and was attempted in [this closed PR](https://github.com/Kitware/CMake/pull/146)), but as can be seen there it does cause quite a bit of structural changes.

However just calling the hook will cover the needs of 99% of the Emacs-users and requires almost no changes at all.

So that's what this patch does, but only after checking that the mode-hook exists to avoid compatibility issues with older versions of Emacs.

### DIFF
--- a/Auxiliary/cmake-mode.el
+++ b/Auxiliary/cmake-mode.el
@@ -259,7 +259,9 @@ the indentation.  Otherwise it retains the same position on the line"
   (setq comment-start "#")
 
   ; Run user hooks.
-  (run-hooks 'cmake-mode-hook))
+  (if (boundp 'prog-mode-hook)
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  Applied as commit b24b58bb7b100794846786521f1977f048ecf5b6.

PR #161, opened 2015-06-23
FindIce: Update for version 3.6.0 and find all Ice programs

rleigh-dundee:
ZeroC will release Ice version 3.6.0 in a week or so.  The FindIce module has been updated to find version 3.6.0 and tested against 3.6.0-rc2 to make sure the logic is correct.

Ice provides a number of daemons and other utility programs; I've also updated the logic to find these for use by end users (since they are visual-studio-version- and architecture-specific).

### DIFF
--- a/Modules/FindIce.cmake
+++ b/Modules/FindIce.cmake
@@ -20,7 +20,7 @@
 #   Ice_SLICE_DIRS - the directories containing the Ice slice interface
 #                    definitions
 #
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  I've merged to `next` for testing per our normal workflow, and will later merge to `master`.  I'll close this for now and re-open if there are problems discovered by testing.

PR #164, opened 2015-07-08
FindXercesC: Find versioned library on Windows

rleigh-dundee:
The MSVC solution/project files provided by upstream add a `_3` suffix to the `.lib` files, and this is also present in the binary builds packaged for Windows.  Check for this in addition to the regular name.


bradking:
Thanks.  This has been merged to `next` for testing and will later be merged to `master`.

PR #170, opened 2015-07-23
Emacs mode: indentation, font-lock, and mode definition

roysc:
This is amended from Kitware/CMake#146, which was closed due to indentation bugs.
* Defines regexps with `rx` comprehensions to be more readable and reusable
* Cleans up `cmake-indent` with more idiomatic Lisp
* Adds font-lock to highlight variables and keywords
* Renames `unscreamify-cmake-buffer` to `cmake-unscreamify-buffer` for consistency
* Defines mode with `define-major-mode` rather than as function


bradking:
Thanks for updating this.  However, the indentation is still broken:

    function(f)
      if(1)
      endif()
    if(0)
    endif()
    endfunction()

roysc:
Okay, I think I've finally fixed indentation so it matches the original behavior.

bradking:
Thanks.  This looks good now.  I rebased the topic, tweaked the commit messages, and merged.  See commit 5593f28fac6d2aeb909370c4daaa19f419e5d596 for the new head of the topic.

PR #171, opened 2015-07-25
Don't mix release and debug ZLIB libraries

mikedld:
By searching for release and debug libraries separately, prevent possible
issues linking debug code to release library or vice versa. Important on
(at least) Windows where CRT is still split in two.

### DIFF
--- a/Modules/FindZLIB.cmake
+++ b/Modules/FindZLIB.cmake
@@ -74,15 +74,17 @@ set(_ZLIB_SEARCH_NORMAL
   )
 list(APPEND _ZLIB_SEARCHES _ZLIB_SEARCH_NORMAL)
 
-set(ZLIB_NAMES z zlib zdll zlib1 zlibd zlibd1)
-
 # Try each search configuration.
 foreach(search ${_ZLIB_SEARCHES})
-  find_path(ZLIB_INCLUDE_DIR NAMES zlib.h        ${${search}} PATH_SUFFIXES include)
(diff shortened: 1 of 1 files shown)


mikedld:
Same problem is in FindPNG (and who knows where else)... Should I make a cumulative pull request or keep it simple and fix one by one?

bradking:
Thanks for working on this.  Many find modules predate the `select_library_configurations` helper and don't have separate per-config cache entries.  This looks like a good start for fixing `FindZLIB` but there are a couple remaining problems:

* We need to remain compatible with pre-populated cache entries like `-DZLIB_LIBRARY=/path/to/libz.a`.
* The `ZLIB::ZLIB` imported target needs to be updated to provide both configurations.

Please see `CONTRIBUTING.rst` for the preferred contribution path.  The dev mailing list will have a broader audience.  That would be a good place to discuss goals for fixing this issue for more modules in general.

mikedld:
Thanks for your reply, Brad. Yes, this commit was a result of me noticing that at least FindZLIB and FindCURL (those are just the ones I'm currently using) don't do well with debug/release libraries. There's also a way for user to override `CMAKE_DEBUG_POSTFIX` when building one or another library, which makes things even more complicated (not sure if this is taken into account in any way, probably not). I'll take a look at `CONTRIBUTING.rst` when time permits :)

bradking:
Closing for now.  Please raise this on the mailing list when ready.

PR #174, opened 2015-07-30
GenerateExportHeader: Add EXPORT_TEMPLATE macro

rleigh-dundee:
In addition to the foo_EXPORT macro, a foo_EXPORT_TEMPLATE macro is also generated.  This is to allow for explicit export of templates in Windows DLLs following Microsoft's guidance here: https://support.microsoft.com/en-us/kb/168958

I'm unaware if there were reasons for this not already being included in the export header, but it appears necessary to properly export templates, and without reinventing the logic in the header to define your own equivalents.  Should be pretty safe to add since it won't affect existing users of GenerateExportHeader.

### DIFF
--- a/Modules/exportheader.cmake.in
+++ b/Modules/exportheader.cmake.in
@@ -4,15 +4,18 @@
 
 #ifdef @STATIC_DEFINE@
 #  define @EXPORT_MACRO_NAME@
+#  define @EXPORT_MACRO_NAME@_TEMPLATE
 #  define @NO_EXPORT_MACRO_NAME@
 #else
 #  ifndef @EXPORT_MACRO_NAME@
 #    ifdef @EXPORT_IMPORT_CONDITION@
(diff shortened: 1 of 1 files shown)


bradking:
Interesting.  IIRC extern templates are part of the C++11 standard, so this is not MS-specific.  I'd like to see some examples of how the two macros may be used together to dllexport an explicit template instantiation and dllimport it elsewhere.  However, I'd prefer to discuss this with a broader audience.

Please raise this for discussion on the [CMake Developers Mailing List](http://www.cmake.org/mailman/listinfo/cmake-developers), which is the preferred contribution path descried in `CONTRIBUTING.rst` anyway.

bradking:
Thanks for starting the thread.  For reference, it is [here](http://thread.gmane.org/gmane.comp.programming.tools.cmake.devel/13854).

PR #177, opened 2015-08-12
Capitalize first alphabet of all MESSAGEs

cicku:
This patch will make CMake output be more formal.


bradking:
Thanks.  Rebased on `master` and applied as commit 50c9f521e2debce5beb538050fd12d802964a4b1.

PR #178, opened 2015-08-13
Emacs mode: font-lock regexps

roysc:
Fixes regular expression for function names, to allow a space between the name and parenthesis.

Adds `- . + /` to variable name regexp, and removes `$ *`.
Removing `$` fixes font-lock in nested substitutions: e.g. in `${${FOO}}` only the second `$` was highlighted; now `FOO` is. Otherwise behavior is the same.


bradking:
Thanks.  I noticed that this breaks variables like `${A_B_C}`.  Only part before the first `_` gets colored.

roysc:
Fixed. `"_-+"` was interpreted as a character range.

bradking:
Thanks.  Applied as commit 758664b7ad018aa19072fe5c359a702c472faae2 and commit 64e6bc5ca3bf15239829f18a71511df2f04181b3.

PR #179, opened 2015-08-17
Add support for HiDPI (Retina) screens in Qt GUI.

Franky47:
Proposing minor change to add HiDPI (Retina) support to the Qt GUI.

It is activated by setting `NSHighResolutionCapable` to `true` in the Info.plist of the application.

Here are two screenshots showing the Qt GUI before and after the change:
before
after


bradking:
Thanks.  Applied as commit a41ebd4ee5b01c92958a28ca2e57b2b6446bf334 with minor tweaks to the commit message.

PR #181, opened 2015-08-28
Add support for setting AppxBundle policy.

thedmd:


bradking:
Thanks!  Please find a place in the test suite to cover this.  Please also read `CONTRIBUTING.rst`.

bradking:
@thedmd, please see my previous comment about adding a test case.

Cc: @khouzam

khouzam:
@thedm thanks for the change. This should not be restricted though to WindowsStore only as it also works for WindowsPhone on 8.1.

bradking:
Closing for now.  Please also read `CONTRIBUTING.rst` and send an updated patch to the mailing list after revising it based on comments here.

PR #182, opened 2015-08-28
Add support for RESW files to VS generator.

thedmd:
Add support for RESW files to Visual Studio 2010+ generatos. Resw files
are added as PRIResource element.

### DIFF
--- a/Source/cmGeneratorTarget.cxx
+++ b/Source/cmGeneratorTarget.cxx
@@ -61,6 +61,7 @@ struct HeaderSourcesTag {};
 struct ExternalObjectsTag {};
 struct IDLSourcesTag {};
 struct ResxTag {};
+struct ReswTag {};
 struct ModuleDefinitionFileTag {};
 struct AppManifestTag{};
 struct CertificatesTag{};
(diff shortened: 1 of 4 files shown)


bradking:
Thanks!  Please find a place in the test suite to cover this.  Please also read `CONTRIBUTING.rst`.

bradking:
@thedmd, please see my previous comment about adding a test case.

Cc: @khouzam

khouzam:
Thanks for adding this support, I'm actually addressing a similar issue with my changes to enable Windows 10 support but in a different way. Because there always seems to have new file types that people want to add which are fairly easy to support generically, I've added a new source property VS_TOOL_OVERRIDE. By default each extension that the VisualStudio generator knows about will appear with their appropriate tool, or none, if the generator doesn't know about the filetype. If you set VS_TOOL_OVERRIDE then you can specify a different tool to apply to that source file. For example, I'm applying the PRIResource tool to a .resw file achieving similar results to what you are doing but with the added advantage that this can add a new file type without needing to teach CMake about it.

RESW files are basic enough that having them natively supported is good, but perhaps the generic way makes it easier to extend.

You can take a look at the change here:
https://github.com/Microsoft/CMake/commit/de374e5969327bc5e1c6f2f9c6da40dc762b1de7

bradking:
Closing for now.  Please also read `CONTRIBUTING.rst` and send an updated patch to the mailing list after revising it based on comments here.

PR #185, opened 2015-09-25
FindGIT convenience functions

rondiplomatico:
added get_git_revision and get_git_branch commands as follow-up to cmake@cmake.org / http://public.kitware.com/pipermail/cmake/2015-September/061516.html
also sent a patch to the devlist.

### DIFF
--- a/Modules/FindGit.cmake
+++ b/Modules/FindGit.cmake
@@ -20,6 +20,24 @@
 #    if(GIT_FOUND)
 #      message("git found: ${GIT_EXECUTABLE}")
 #    endif()
+#
+# For convenience, the package also provides the following functions:
+#
+# ::
+#
+#    git_get_revision(VARNAME [WORKING_DIRECTORY])
+#    git_get_branch(VARNAME [WORKING_DIRECTORY])
+#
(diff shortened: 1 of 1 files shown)


bradking:
Thanks for working on this.  As explained in `CONTRIBUTING.rst` and in [a response](http://thread.gmane.org/gmane.comp.programming.tools.cmake.user/53415/focus=53436) to the thread you linked, sending a patch to the developer list is the preferred contribution path.  I do not see any post about this patch on the dev list.  Did you sign up for the list?  It is member-post-only.

Please also see the [review comment here](http://thread.gmane.org/gmane.comp.programming.tools.cmake.user/53415/focus=53439) about using `.. comand::` in the documentation.

Instead of using `rev-list` to get the version, one could use `git rev-parse --verify -q HEAD`.  This is what CMake does in its own `Source/CMakeVersionSource.cmake`.

bradking:
Okay, the post to the dev list has now appeared [here](http://thread.gmane.org/gmane.comp.programming.tools.cmake.devel/14560).  We'll pick up discussion over there so I'm closing this here.

PR #186, opened 2015-10-01
FindOpenSSL Module - MSVC Runtime switching support

thoys:
FindOpenSSL Module - MSVC Runtime switching support (/MD default , /MT when you switch OPENSSL_MSVC_SHARED_RT off)

### DIFF
--- a/Modules/FindOpenSSL.cmake
+++ b/Modules/FindOpenSSL.cmake
@@ -37,6 +37,7 @@
 #
 # Set ``OPENSSL_ROOT_DIR`` to the root directory of an OpenSSL installation.
 # Set ``OPENSSL_USE_STATIC_LIBS`` to ``TRUE`` to look for static libraries.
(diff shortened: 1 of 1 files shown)


thoys:
@bradking  Came across this when I found out many of our users didn't have the MSVC Runtime installed. This makes it easier to make a build for users without administrator permissions. I am unable to set the startup directory of my application so I will link everything I require in the executable.

bradking:
Thanks for working on this.  Since shared RT is the default I think I'd rather name the option `OPENSSL_MSVC_STATIC_RT`, drop the explicit default code, and invert the semantics.

Please also read `CONTRIBUTING.rst` for the preferred contribution path.

bradking:
Thanks for the update.  I've applied the change with minor tweaks as commit 821e17785ecd4a329842dc6c5b093eb30cc4f160.  Please verify that this version works for you.

PR #188, opened 2015-10-04
Document and test CMAKE_[CURRENT_](BINARY|SOURCE)_DIR in script mode

tamaskenez:
Reference thread: http://public.kitware.com/pipermail/cmake-developers/2015-October/026612.html

This commit updates only documentation and tests, that's why I'm pushing to next.
Let me know if I should rebase to master instead.


bradking:
Thanks.  Please rebase on `master` (`next` is rewritten every 4 months or so, and happened to be this morning).  Also, please simply post the patch to the mailing list right in the thread you linked.  This is the preferred path for patches as mentioned in `CONTRIBUTING.rst`.

tamaskenez:
Okay, thanks, I saw a lot of PRs here, that's why I thought I could post here.
I rebased to master and sent it to mailing list.

PR #190, opened 2015-10-08
cmake-mode: treat keywords as symbols

kcburge:
Symbols such as "CHECK_FUNCTION_EXISTS" should not trigger an open
block (due to containing "FUNCTION").


bradking:
Thanks.  It looks like this was recently regressed.  Fix applied as commit a935745d0730cf1cd6c1a880d71e36a39ad86f8c.

PR #192, opened 2015-10-08
cmake-mode.el: unscreamify symbols instead of words

kcburge:


kcburge:
Does not work on all CMakeLists.txt files.

PR #193, opened 2015-10-08
cmake-mode.el: unscreamify symbols instead of words

kcburge:


bradking:
Thanks.  Applied as commit 47f4b3b662da7f49fd0353b978eaf781608e9282.

PR #194, opened 2015-10-12
FindGit: add support for SourceTree on Windows

kainjow:
Fixes Mantis bug [#15758](https://public.kitware.com/Bug/view.php?id=15758)


bradking:
Thanks.  Applied with minor tweaks as commit 57f701a7f045fc13807933860bad22fa405c6621.  Please verify that it still works for you.

PR #195, opened 2015-10-13
Add find_package() Support for PostgreSQL 9.5

mtangoo:
I just found that PGS 9.5 isn't supported. this PR add just that. Its tested


bradking:
Thanks!  This change happened to be made already by commit 5225e0048c2b734eba7b03e396f89e74214daa66 yesterday based on a patch sent to our dev mailing list.

mtangoo:
Didn't notice it :)

PR #196, opened 2015-10-19
Add support for multilingual SLAs for the CPack DragnDrop Generator

sonOfRa:
Multiple languages for SLAs and the SLA UI can be added via the CPack variables
CPACK_DMG_SLA_DIR and CPACK_DMG_SLA_LANGUAGES. For each language defined in the
languages variable, CPack will search for <language>.menu.txt and
<language>.license.txt in CPACK_DMG_SLA_DIR. If the sla directory variable is not
defined, the old behaviour using CPACK_RESOURCE_FILE_LICENSE is retained.

This adds a library dependency on the deprecated Carbon Framework to CPackLib, since the functions/types in https://github.com/Kitware/CMake/compare/master...sonOfRa:multi-sla?expand=1#diff-9cd479a075495d596d521ca1d4575b99R554 are needed for the internal LPic data block in the SLA. Apple deprecated these APIs without offering a replacement.

Another thought concerning the ```<language>.menu.txt``` files: It might be out of scope, but do you think these files could be added to the CMake repository for "common" languages? English is hardcoded as a string in the source file already. I'd suggest the following behaviour: If an internal version exists, use that, unless the file ```CPACK_DMG_SLA_DIR/<language>.menu.txt``` is found, allowing the user to override the internal version. If no internal version exists, error, unless ```CPACK_DMG_SLA_DIR/<language>.menu.txt``` is found.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -728,6 +728,10 @@ endif()
 # Build CPackLib
 add_library(CPackLib ${CPACK_SRCS})
 target_link_libraries(CPackLib CMakeLib)
+if(APPLE)
+  include_directories (${CMAKE_OSX_SYSROOT}/System/Library/Frameworks/Carbon.framework/Headers/)
+  target_link_libraries(CPackLib "-framework Carbon")
+endif()
 
 if(APPLE)
   add_executable(cmakexbuild cmakexbuild.cxx)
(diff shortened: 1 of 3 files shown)


bradking:
Thanks for working on this.  These changes touch on areas of OS X development I'm not familiar with so I'd prefer to have this discussed with a broader audience.  Please read `CONTRIBUTING.rst` and post this patch to the dev mailing list.

PR #199, opened 2015-10-31
Change FindPythonLibs to ask the interpreter for its config (master).

NikolausDemmel:
This askes the python interpreter what the location of libs and headers is instead
of relying on finding the libraries/headers on the file system.


NikolausDemmel:
meant to open this in my fork. sorry for the noise.

PR #200, opened 2015-11-25
Fix off-by-one when adding two times

AMDmi3:
Microseconds range from 0 to 999999, so 1000000 is an overflow which should be processed as well.


bradking:
Thanks.  I ported this to upstream KWSys here:

 http://review.source.kitware.com/20444

Once integrated there it will be merged to CMake.

AMDmi3:
Thank you!

PR #201, opened 2015-11-29
FindOpenAL: Fix Windows architecture detection error

Zogzer:
Certian OpenAL implementations on Windows provide both Win32 and Win64 binaries, the bug caused Win32 libraries to be used even if the project was 64bit, this fixes that issue by checking the project architecture and only providing the correct library location


bradking:
Thanks.  Please revise the commit message to start in `FindOpenAL: `.  Also, please add a surname to the Git authorship record.

bradking:
Thanks.  Applied as commit 39abbaed7756ac9b4d1d2e7f44575ae85388aaf6.

PR #202, opened 2015-12-02
CMakeParseArguments: Test suite and native implementation

metti:
This pull request adds a test suite for the CMakeParseArguments module. In addition the second change implements cmake_parse_arguments(...) as native command.

In our project we saw that cmake_parse_arguments is actually called a lot of times due to macro expansion and we could almost halve our configure time. This implementation follows the implementation of the former macro and both implementations of course pass the added test suite.

Arguably one could implement a warning mechanism (and tests accordingly) for options/etc. that are specified more than once. 
E.g. cmake_parse_arguments(prefix "OPT" "OPT" "" OPT foo). To me it appears a bit random what the outcome of this usage shall be (either prefix_OPT is TRUE or prefix_OPT is 'foo'). A -Wdev warning could be helpful to eliminate this ambiguity, but might need the introduction of a policy. Let me know whether I should add this to strengthen the implementation.


bradking:
Great start!  I've long wanted to see this done.  Here are some comments.

* The `Tests/CMakeTests` infrastructure has been superseded by the `Tests/RunCMake` infrastructure.  Please add a `Tests/RunCMake/cmake_parse_arguments` test to cover the command instead.  See `Tests/RunCMake/README.rst` for help.  This approach will also allow the error cases to be covered.

* Please revise the documentation as well.  The old module should be documented as deprecated.  The new command should get a `Help/command/cmake_parse_arguments.rst` file to document it.

metti:
Besides addressing your comments I added a third change that makes cmake_parse_arguments() emit an error in case keywords are specified more than once. This actually would break compatibility (which is why I made it a separate commit) for some probably unused corner cases. But in my opinion this improves the implementation as I assume that most of the duplicate specifications are actual errors that else get silently ignored.

Let me know if you have any further comments to address.

bradking:
Thanks.  I agree it is nicer to diagnose duplicates but for compatibility we cannot make it an error now.  Please make it a warning for now instead.

We could consider introducing a policy to make it an error as a follow-up to this PR.  To do that, please read `CONTRIBUTING.rst` and come to the mailing list to discuss the possibility.

bradking:
Thanks for bringing this over [to the dev list](http://thread.gmane.org/gmane.comp.programming.tools.cmake.devel/15273).  Closing here since we'll pick up review over there.

PR #203, opened 2015-12-09
Allow GRAPHVIZ_GRAPH_NAME to contain spaces without requiring escaped quotes.

mishchea:
Without this patch, `SET (GRAPHVIZ_GRAPH_NAME "hello world")` does not work (it results in a parsing error in GraphViz when the generated output is processed), but `SET (GRAPHVIZ_GRAPH_NAME "\"hello world\"")` does.


bradking:
Thanks!  Applied as commit e0ad72d8af9bfba80d24f823d79440ebb9136bda.

PR #204, opened 2015-12-11
Use correct capitalization for Protobuf

sschuberth:
All other modules use their module name (e.g. XxX for FindXxX.cmake) in
find_package_handle_standard_args. Protobuf used all-caps, which triggers
a bug when we try to find Protobuf with the CMakeFindDependencyMacro.cmake
macro, which only checks for the mixed-case _FOUND.


bradking:
Thanks.  Applied as commit 2908103d04daeaf396173b6f868f2cad5eb47b08.

PR #206, opened 2016-01-15
Minor Ninja generator fixes for Windows

sschuberth:


sschuberth:
@bradking Would you mind also having a look at this?

bradking:
Thanks.  However, the `&` change breaks some tests, such as `RunCMake.CMP0060`.  Please try running the full test suite.  Build CMake itself using the Ninja generator and then run `ctest -j 4` at the top of the build tree.

Are you sure about `&` v. `&&`?  Can you link to documentation about it?

sschuberth:
Looking at the [official documentation](https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/ntcmds_shelloverview.mspx?mfr=true) again I see why there would be test failures: While it's true that `&` is used to separate multiple commands, it executes the next command no matter whether the previous was successful or not. I.e. it corresponds to Unix' `;`. To execute the next command only on success of the previous one still `&&` needs to be used. Sorry for the confusion.

As the other commit in my PR is not important enough on its own, I'll simply close this. Thanks again for the thorough review.

PR #207, opened 2016-01-15
Restore of PKG_CONFIG_PATH should use _pkg_config_path_old

codesmythe:
When restoring the old value of the PKG_CONFIG_PATH environment variable, I think _pkgconfig_path_old should be used.


bradking:
Good catch, thanks.  Applied with revised commit message as commit 8979a1070097fde3ca0ab45bdd6580b7acd07ca4.

PR #208, opened 2016-01-16
Put ${_pkgconfig_path} in quotes to protect semicolons on Windows.

codesmythe:
Consider the case on Windows where the pkgconfig_path is multiple directories separated by semicolons (.e.g, "c:\foo;c:\bar;c:\baz"). In the existing code, this path would be interpreted as list and ENV{PKG_CONFIG_PATH} would only get the first item in the path.


bradking:
Thanks.  I rebased on `master` after merging #207 so we can quote the old value too, just in case.  Please see commit 53511936e43c14f362b1e1ba069a77f5ba251ccc.

PR #209, opened 2016-01-18
Allow users of CheckCXXSourceRuns and CheckCSourceRuns to access the …

hm1992:
…output from the run

While making custom finders for 3rd party libraries I was faced with some that didn't allow the version to be stripped from the header. To get around that I can use CheckCXXSourceRuns to make and run a small .cpp file which prints the version to stdout. This allows me to grab that using CheckCXXSourceRuns rather than try_run which is more difficult to use.


bradking:
Thanks for also submitting this to the cmake dev list [here](http://thread.gmane.org/gmane.comp.programming.tools.cmake.devel/15502).  That is currently the preferred place for discussion so I'll close this here and pick it up over there.

PR #211, opened 2016-01-22
sort buildPaths

exmatrikulator:
If you have a lot of projects with different configurations, it's hard to find your correct buildPath.
Now it should be easier ;)


bradking:
Currently the order is from most recently used to least recently used.  This is intended to be a "working set" that changes with usage.  The length is capped at 10 so the list never gets too long.  The proposed patch appears to sort the order only for display in the drop-down so the working set update is unchanged.

Cc: @clintonstimpson

bradking:
Since there are at most 10 items on the list it is not very hard to find the right one regardless of the order.  The current sort by recency is convenient for switching among a small number of build trees.  Therefore I do not think we should make this change.  The [cmake developer list](https://cmake.org/mailman/listinfo/cmake-developers) would be a better place to discuss this so I'm closing this PR here.

PR #212, opened 2016-01-30
IAR compiler flags for creating working Ninja build files

osimola:
The dependency flags require recent versions of iccarm and iccavr. The
multi-rule dependency generated with --dependencies=m does not work well
with Ninja.


bradking:
Thanks.  Applied as commit 99a9c7e51a3980807c60a9ec7532db0ca9495e5f.

I don't know what `--dependencies=m` generates, but [Ninja PR 1039](https://github.com/ninja-build/ninja/pull/1039) may be related.

osimola:
Thanks. 

That PR is indeed related to the same problem. --dependencies=n generates the dependency files in the first format, single target for every dependency (ns skips system headers). It's only present in newer versions of the IAR compilers, from somewhere around 2012 if I recall correctly. --dependencies=m generates the second format, which doesn't work with Ninja yet.

PR #214, opened 2016-02-04
Error out if --target is specified multiple times

sschuberth:
So far it was not clearly documented that --target is only supported to be
specified once. Even worse, specifying it multiple times would silently
ignore any previously specified targets and only build the last target.

Improve the docs in this regard and error out if --target is specified
more than once in order to prevent user errors.


sschuberth:
@robertoraggi @haraldF Mind having a look?

bradking:
Thanks.  Please extend `Tests/RunCMake/CommandLine/RunCMakeTest.cmake` with a test case covering this.  It already has a few cases covering `--build` errors.

sschuberth:
I'll happily write a test. Is there some documentation on how to use `run_cmake_command` when expecting failures, or how to capture and compare *stderr* output?

bradking:
> Is there some documentation on how to use run_cmake_command

See `Tests/RunCMake/README.rst` for a general introduction.  You need to create a `-result.txt` and `-stderr.txt` containing the expected return code (e.g. `1`) and a regex matching the expected stderr, respectively.  See the `build-no-dir` case for an example.

sschuberth:
I've pushed a work-in-progress commit that adds a test. However, I'd need some more help with that. Currently, I only added a dummy for the expected *stderr* output, but still the test passes for me. What's the correct command to only run the *CommandLine* test after I've modified `CommandLine/RunCMakeTest.cmake`?

Tests/RunCMake/CommandLine/RunCMakeTest.cmake
```diff
@@ -54,6 +54,9 @@ function(run_BuildDir)
 endfunction()
 run_BuildDir()
 
+run_cmake_command(build-multiple-targets ${CMAKE_COMMAND} -E chdir ..
+  ${CMAKE_COMMAND} --build BuildDir-build --target CustomTarget --target CustomTarget2)
+
```
bradking: Put these two lines at the bottom of the `run_BuildDir` function so that they run in the context of that test.

sschuberth:
I believe to have fixed the test now. For the record, I was looking at the wrong test results (those for *CommandLineTest* instead of *RunCMake.CommandLine*) at first.

Now I successfully ran only the latter test via

    $ ctest -R "^RunCMake.CommandLine$"

bradking:
Thanks.  The test looks good.  While reading the first commit again I realized we should do the check when we see `--target` again and immediately fail.  The option should not be allowed twice regardless of whether it has a value the second time.  Also, please update the error message to mention the `--target` option by name.

sschuberth:
Thanks for the feedback. I'm not sure I agree on the proposed behavior, though. I just realized that it's indeed valid to run `cmake` without `<tgt>` being specified for `--target`, i.e. like

    cmake --build . --target

That by itself is a bug IMO. `cmake` should check for mandatory arguments to be present. Anyway, let's assume for now an empty `<tgt>` stands for "default target". If so, doing something like

    cmake --build . --target --target

would not actually be a problem, and neither would

    cmake --build . --target some_target_name --target some_target_name

be as building the same target twice should make the second invocation a no-op. So I think we could allow this. Do I get you correctly in that you're saying that you do *not* want to allow the same target being specified multiple times, and even pathological cases like 

    cmake --build . --target --target

should fail? I acknowledge that at least that would be consistent with the docs saying that `--target` may appear only once (no matter what `<tgt>` is).

bradking:
The intention of `--target` was always to be `--target <tgt>` but the current command-line parsing is simply not enforcing this.  Using `--target --target` could be interpreted as asking for `<tgt> == --target`.  Ideally we should clean up the `--build` argument parsing to be more strict for all the options.

sschuberth:
Alright, I've force-pushed an update. Mind having another look?

bradking:
Thanks.  Applied as commit 497cad7c883dc401b4d78109c3a057fb38745d9e and merged to 'next' for testing.  Once cleared there it will be merged to 'master'.

PR #215, opened 2016-02-08
Add '%s' (UNIX time_t format) to command STRING(TIMESTAMP ...)

jlblancoc:
This may be useful when implementing SOURCE_DATE_EPOCH in a
cross-platform way.
See: https://reproducible-builds.org/specs/source-date-epoch/

[Docs online](https://cmake.org/cmake/help/v3.0/command/string.html) should be updated, adding this format specifier: 

%X       The numerical value of the UNIX (time_t) timestamp.


bradking:
The `%X` placeholder appears to be commonly used for a different purpose.

According to `man date`:

    %X     locale's time representation (e.g., 23:13:48)

According to `man strptime`:

    %X     The time, using the locale's time format.

Perhaps we need a `%s` extension:

    %s    The number of seconds since the Epoch

The current implementation offers the same placeholders as MS [strftime](https://msdn.microsoft.com/en-us/library/fe06s4ak.aspx) on Windows so any new ones need custom implementations on at least one platform.  We would also need to account for any difference in Epoch choice across platforms.

jlblancoc:
Good point! Just switched it to '%s' to make it consistent with `date`. 

However, please note that the code already deals with the '%s' flag in a special code block which is independent of the platform, so it will work on Windows and Linux. See [this line](https://github.com/Kitware/CMake/pull/215/files#diff-0dea18e51a7ed4baf7eccbbf837f60dfR88).

bradking:
Thanks.  It does look like [MS time](https://msdn.microsoft.com/en-us/library/1f4c8f33.aspx) uses the same Epoch according to its documentation.  I verified with local tests also.

Please move the implementation into a switch case in `cmTimestamp::AddTimestampComponent`.  Please also update the documentation in `Help/command/string.rst` and extend the test suite (grep for "TIMESTAMP" in the Tests directory to see current tests).

jlblancoc:
@bradking done!

bradking:
@ngladitz please take a look at this since you added cmTimestamp originally.

ngladitz:
The C standard says that it neither defines precision nor range for time_t.

CMake supports more platforms / toolsets than I am familiar with so there may or may not be one or more among them where the reference point isn't January 1, 1970 or where it isn't second precision.

I wouldn't feel entirely comfortable with this unless we could confirm (with a test) that this holds true for all the supported implementations.

bradking:
> The C standard says that it neither defines precision nor range for time_t.

Good point.  Is there a POSIX (and Windows) API we can use to convert it officially to number of seconds since the epoch (even if in practice it would be a no-op)?

ngladitz:
For GNU there is timegm().
The manual page has a listing for a "portable" [1] alternative.
For windows (MSVC) there is _mkgmtime().

With those we could take a struct tm with "January 1, 1970" and convert it to time_t which would then represent unix epoch.

Then we could get the difference between the given time_t and unix epoch with difftime() which is defined by the standard to be in seconds.

P.S. Random implementation I found where the reference point isn't January 1, 1970:
http://www.nongnu.org/avr-libc/user-manual/group__avr__time.html#ga3346b04b0420b32ccf6b706551b70762

[1] http://man7.org/linux/man-pages/man3/timegm.3.html

jlblancoc:
And I though that I worried about portability... :-) 

I'm making those changes, will update when they are tested in Windows & GNU/Linux. 

What I can't find is a clear way to add unit tests for this functionality, since it depends on the machine time so the outcome is not deterministic... Hope the manual testing is enough for accepting this PR...

jlblancoc:
The portable version in the [man page](http://man7.org/linux/man-pages/man3/timegm.3.html) is now included. 
The code is now tested to build and `%s` to generate the expected value under Windows and GNU/Linux.

ngladitz:
The code also currently does not handle e.g. MinGW which I think uses an old version of the msvcrt but does not define _MSC_VER and from what I can tell may not provide _mkgmtime() [1].

If the "portable" man page provided implementation works on windows (without _mkgmtime()) we could use that throughout(?).

[1] http://mingw.5.n7.nabble.com/Missing-mkgmtime-in-time-h-td8976.html

jlblancoc:
I could test it... what version(s) of MinGW should be tested??

ngladitz:
I don't think the version of MinGW matters ... they likely all use the same old MSVCRT(?)

jlblancoc:
It's done now. CMake is tested to build (and the new command to work correctly) under MSVC, GCC and MinGW.

ngladitz:
Otherwise this looks good to me. Good work!

jlblancoc:
Thanks! For me it's done for now...

bradking:
After some revisions to fix issues revealed by our nightly testing, this has been applied as commit 
6727270b75eec4e379acd4a39f5003a316249c73.

PR #216, opened 2016-02-17
[PATCH] Create DEBIAN directory for dpkg-shlibdeps

apollo13:
dpkg-shlibdeps searches for the DEBIAN directory in order to
resolve $ORIGIN in the rpath to a directory. We need to
create the DEBIAN directory for this to work.

Fixes 12431


bradking:
Cc: @do-m-en

do-m-en:
Thanks, patch applied with some additions:
https://cmake.org/gitweb?p=cmake.git;a=commit;h=b8263a2

PR #217, opened 2016-02-22
Imported working dependency fix from hughperkins

borisfom:
It appears that simple dependency fix was not enough for Torch, we still kept rebuilding all .cu files. I have merged his changes here.


borisfom:
The dependencies were actually working with CMake lates code; closing the request.

borisfom:
Apparently, the latest source with your fixes worked - I withdraw this request.

PR #218, opened 2016-02-26
Android/NVPACK detection fixes, new helper file

borisfom:
Here, made it working for Tegra NDK. 
Refactored cross-compilation section.
Added select_compute_arch helper module to facilitate setting nvcc arch parameters.
Some outdated checks and settings cleaned up.


Modules/FindCUDA.cmake
```diff
@@ -439,6 +438,8 @@ endmacro()
 ###############################################################################
 ###############################################################################
 
+option(CUDA_USE_STATIC_CUDA_RUNTIME "Use the static version of the CUDA runtime library if available" ON)
```
jamesbigler: You removed the code that only enabled this for when the static runtime library was present.  I try to make this work for a wide range of CUDA versions.

Modules/FindCUDA.cmake
```diff
@@ -586,6 +587,49 @@ if(NOT "${CUDA_SDK_ROOT_DIR}" STREQUAL "${CUDA_SDK_ROOT_DIR_INTERNAL}")
...
+
+  if (EXISTS "${CUDA_TOOLKIT_ROOT}/targets/${CUDA_TOOLKIT_TARGET_NAME}")
+    set(CUDA_TOOLKIT_TARGET_DIR "${CUDA_TOOLKIT_ROOT}/targets/${CUDA_TOOLKIT_TARGET_NAME}" CACHE PATH "CUDA Toolkit target location.")
+    SET (CUDA_TOOLKIT_ROOT_DIR ${CUDA_TOOLKIT_ROOT})
+    mark_as_advanced(CUDA_TOOLKIT_TARGET_DIR)
+  endif()
+
+  # Target CPU architecture: this nvcc option is now obsolete
+  # add known CUDA targetr root path to the set of directories we search for programs, libraries and headers
+  set( CMAKE_FIND_ROOT_PATH "${CUDA_TOOLKIT_TARGET_DIR};${CMAKE_FIND_ROOT_PATH}")
```
jamesbigler: This is kind of anti-social.  You should be adding the CUDA_TOOLKIT_TARGET_DIR to the individual find_* calls inside here rather than adding it to a global search path.

Modules/FindCUDA.cmake
```diff
@@ -593,12 +637,14 @@ if(NOT CUDA_TOOLKIT_ROOT_DIR)
   find_path(CUDA_TOOLKIT_ROOT_DIR
     NAMES nvcc nvcc.exe
     PATHS
+      ENV CUDA_TOOLKIT_ROOT
```
jamesbigler: Where does this environment variable come from?

Modules/FindCUDA.cmake
```diff
@@ -611,7 +657,9 @@ if(NOT CUDA_TOOLKIT_ROOT_DIR)
     string(REGEX REPLACE "[/\\\\]?bin[64]*[/\\\\]?$" "" CUDA_TOOLKIT_ROOT_DIR ${CUDA_TOOLKIT_ROOT_DIR})
     # We need to force this back into the cache.
     set(CUDA_TOOLKIT_ROOT_DIR ${CUDA_TOOLKIT_ROOT_DIR} CACHE PATH "Toolkit location." FORCE)
+    set(CUDA_TOOLKIT_TARGET_DIR ${CUDA_TOOLKIT_ROOT_DIR})
```
jamesbigler: Where's the documentation for CUDA_TOOLKIT_TARGET_DIR?  I think there is one for CUDA_TOOLKIT_ROOD_DIR, so you should have one here.

Modules/FindCUDA.cmake
```diff
@@ -634,6 +682,9 @@ find_program(CUDA_NVCC_EXECUTABLE
 find_program(CUDA_NVCC_EXECUTABLE nvcc)
 mark_as_advanced(CUDA_NVCC_EXECUTABLE)
 
+#MESSAGE("CUDA_NVCC_EXECUTABLE: ${CUDA_NVCC_EXECUTABLE} CUDA_TOOLKIT_ROOT_DIR: ${CUDA_TOOLKIT_ROOT_DIR} CUDA_TOOLKIT_TARGET_DIR: ${CUDA_TOOLKIT_TARGET_DIR}")
```
jamesbigler: Try to remove commented out code from submissions.

Modules/FindCUDA.cmake
jamesbigler: Oh, I see this was already present.

Modules/FindCUDA.cmake
```diff
@@ -734,15 +767,14 @@ if(NOT CUDA_VERSION VERSION_LESS "5.5")
...
-else()
-  option(CUDA_USE_STATIC_CUDA_RUNTIME "Use the static version of the CUDA runtime library if available" OFF)
-endif()
 
-if(CUDA_USE_STATIC_CUDA_RUNTIME)
-  if(UNIX)
+if(NOT CUDA_USE_STATIC_CUDA_RUNTIME STREQUAL OFF)
+  if(CUDA_cudart_static_LIBRARY)
+    # Set whether to use the static cuda runtime.
+    set(CUDA_USE_STATIC_CUDA_RUNTIME ON CACHE BOOL "Use the static version of the CUDA runtime library if available")
```
jamesbigler: You've already set this above, so this call won't have an effect.

Modules/FindCUDA.cmake
```diff
@@ -764,15 +796,17 @@ if(CUDA_USE_STATIC_CUDA_RUNTIME)
...
       if (NOT CUDA_rt_LIBRARY)
         message(WARNING "Expecting to find librt for libcudart_static, but didn't find it.")
       endif()
     endif()
-  endif()
-endif()
+  endif(UNIX AND NOT ANDROID)
+endif(NOT CUDA_USE_STATIC_CUDA_RUNTIME STREQUAL OFF)
+
+# MESSAGE("CUDA_TOOLKIT_INCLUDE: ${CUDA_TOOLKIT_INCLUDE} CUDA_VERSION_STRING: ${CUDA_VERSION_STRIN} CUDA_CUDART_LIBRARY: ${CUDA_CUDART_LIBRARY} CUDA_cudart_static_LIBRARY: ${CUDA_cudart_static_LIBRARY}")
```
jamesbigler: Kill commented out code unless it is documentation.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,176 @@
...
+# Variables affecting the choice:
+#
+# CUDA_ARCH_NAME: One of ("Fermi" "Kepler" "Kepler-M" "Maxwell" "Maxwell-M" "All" "Manual")
+# ENV{CUDA_ARCH_BIN} : set if CUDA_ARCH_NAME set to "Manual"
+#
+
+# Known NVIDIA GPU achitectures
+# This list will be used for CUDA_ARCH_NAME = All option
+
+if (NOT CUDA_arch_select_included)
```
jamesbigler: Why do you need include guards?  This isn't the typical CMake behavior.

Modules/FindCUDA/compute_arch.cmake
```diff
@@ -0,0 +1,152 @@
...
+
+# Known NVIDIA GPU achitectures
+# This list will be used for CUDA_ARCH_NAME = All option
+SET(KNOWN_GPU_ARCHITECTURES "2.0 2.1(2.0) 3.0 3.2 3.5 5.0 5.2 5.3")
+
+################################################################################################
+# Removes duplicates from LIST(s)
+# Usage:
+#   LIST_UNIQUE(<list_variable> [<list_variable>] [...])
+MACRO(LIST_UNIQUE)
```
jamesbigler: I already commented on my feelings about removing this macro in the email I sent.  I think the code is clearer without it.

Modules/FindCUDA/compute_arch.cmake
```diff
@@ -0,0 +1,152 @@
...
+      LIST(REMOVE_DUPLICATES ${__lst})
+    ENDIF()
+  ENDFOREACH()
+ENDMACRO()
+
+################################################################################################
+# A function for automatic detection of GPUs installed  (IF autodetection is enabled)
+# Usage:
+#   DETECT_INSTALLED_GPUS(OUT_VARIABLE)
+FUNCTION(DETECT_INSTALLED_GPUS OUT_VARIABLE)
```
jamesbigler: Again, all macros, variables, and functions in the FindCUDA family need to have CUDA_ prefixed.

Modules/FindCUDA/compute_arch.cmake
```diff
@@ -0,0 +1,152 @@
...
+    SET(__archs_name_default "Auto")
+  ENDIF()
+
+  # SET CUDA_ARCH_NAME strings (so it will be seen as dropbox in CMake-Gui)
+  SET(CUDA_ARCH_NAME ${__archs_name_default} CACHE STRING "Select target NVIDIA GPU achitecture.")
+  SET_property( CACHE CUDA_ARCH_NAME PROPERTY STRINGS "" ${__archs_names} )
+  mark_as_advanced(CUDA_ARCH_NAME)
+
+  # verIFy CUDA_ARCH_NAME value
+  IF(NOT ";${__archs_names};" MATCHES ";${CUDA_ARCH_NAME};")
```
jamesbigler: There are list(FIND) type commands.  Why not use those over string matches?

borisfom: Absolutely - I am a CMake novice so follow safer path to use the same constructs as before :)

Modules/FindCUDA.cmake
borisfom: No that would not work (in fact I tried that first). The scope of find_ is already restricted by CMAKE_FIND_ROOT_PATH. so adding the path that is not under CMAKE_FIND_ROOT_PATH makes no effect.

Modules/FindCUDA.cmake
borisfom: Tegra (NVPACK) sets this one on install.

Modules/FindCUDA.cmake
borisfom: Those are only different in cross-compile mode : TARGET_DIR only has Android binaries while ROOT_DIR is the root of the whole package (including targets).

Modules/FindCUDA.cmake
borisfom: probably I should put those under if (verbose) - having those when needed saves a lot of typing.

Modules/FindCUDA.cmake
borisfom: Well, I found that I could not possibly override it from outside no matter what. Even though it looked like it should have worked. This way override works, also I find it more intuitive for all options decraled upfront. Keeping the option enabled only if we find static lib is still done, further down.

Modules/FindCUDA.cmake
borisfom: Yep, that's documentation - I modified original syntax together with adding the version check as it seemed to have changed. It;s a kinf of 'developer' doc though so I'd rather kill it.

Modules/FindCUDA/compute_arch.cmake
borisfom: Sure will do.

Modules/FindCUDA/compute_arch.cmake
borisfom: Noted for the next iteration

Modules/FindCUDA/select_compute_arch.cmake
borisfom: I found the inclusion of this file via cuda_find_helper_file from FindCUDA makes names like DETECT_INSTALLED_GPUS be hidden from the outside callers, which broke some existing Torch code where I extracted FindCudaArch.cmake from. So for backward compatibility I have provided FindCudaArch.cmake in my repo that includes this file via include(). Since I was not sure what happens if I include it twice from 2 different places I have added those. Probably should be removed, what woudl you say ?

borisfom: The right way to fix it is to export SELECT_NVCC_ARCH_FLAGS only to be visible by FindCUDA clients. I believe those should be left for the user to set and not set automatically.  How do I do that ?

Modules/FindCUDA.cmake
```diff
@@ -734,15 +767,14 @@ if(NOT CUDA_VERSION VERSION_LESS "5.5")
...
-if(CUDA_cudart_static_LIBRARY)
-  # Set whether to use the static cuda runtime.
-  option(CUDA_USE_STATIC_CUDA_RUNTIME "Use the static version of the CUDA runtime library if available" ON)
-else()
-  option(CUDA_USE_STATIC_CUDA_RUNTIME "Use the static version of the CUDA runtime library if available" OFF)
-endif()
 
-if(CUDA_USE_STATIC_CUDA_RUNTIME)
-  if(UNIX)
+if(NOT CUDA_USE_STATIC_CUDA_RUNTIME STREQUAL OFF)
```
jamesbigler: Since CUDA_USE_STATIC_CUDA_RUNTIME is a "BOOL" variable you should do:

if(NOT CUDA_USE_STATIC_CUDA_RUNTIME) or if(CUDA_USE_STATIC_RUNTIME) instead of comparing against a string representation of the value (which could be OFF or 0).

Modules/FindCUDA.cmake
jamesbigler: You should note that the effect of set(VAR value CACHE ...) is only the first invocation ever during configure.  Once it is in the cache, then this call has no effect unless you apply a FORCE argument which will always override the value in the CACHE (despite what the user by want to change via the GUI or some other means).  It is more useful to provide a default value based on the presence of CUDA_cudart_static_LIBRARY like I had before.

Modules/FindCUDA.cmake
jamesbigler: These messages only print during configure time, and don't follow any "verbose" mode that I know of (perhaps there is a new verbose mode).

Really this sort of debugging output belongs in the calling code and not in here.

Modules/FindCUDA.cmake
jamesbigler: Please read the documentation for find_executable/find_library (NO_DEFAULT_PATHS or something like that).  There are ways of adding the CUDA_TOOLKIT_TARGET_DIR as a search path.  I already do this for CUDA_TOOLKIT_ROOT_DIR.  I want to find the executables and libraries in there before I start looking around in default paths.

borisfom: In fact, I read the documentation: https://cmake.org/cmake/help/v3.0/manual/cmake-toolchains.7.html?highlight=cmake_find_root_path

In your case, yes, having CUDA_TOOLKIT_ROOT_DIR in CMAKE_FIND_ROOT_PATH and adding CUDA_TOOLKIT_TARGET_DIR works, since CUDA_TOOLKIT_TARGET_DIR is under CUDA_TOOLKIT_ROOT_DIR Vice versa - does not work unless one changes the setting defined by the toolchain cmake files: 

# only search for libraries and includes in the ndk toolchain                                                                                                  
set( CMAKE_FIND_ROOT_PATH_MODE_PROGRAM ONLY )
set( CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY )
set( CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY )

In NVPACK case, CUDA root is outside of Android sysroot set by toolchain file.
If I change 'ONLY' to both then yes I will be able to find the packages outside the target sysroot, but then I will risk to find host (and other targets)  CUDA library instead of the one I was looking for.
So the right thing to do here is exactly what I did: if target's CUDA library is being installed outside of the known target system "root", add the target CUDA directory to the set of roots the search is being restricted to.

Modules/FindCUDA.cmake
borisfom: I will remove that.

Modules/FindCUDA.cmake
borisfom: Right. I have changed it to strequal first to see if setting the option is not working for the same reason if(X) on macro argument X - but it was not the case. Flipping back to bool.

Modules/FindCUDA.cmake
borisfom: I see resetting cached variable may be not the best idea - I am pretty sure I was not able to turn off static setting on the first run, too, and the option is only useful if the override works...

jamesbigler: Typically if I want to override a default I do one of two things.

If I want to override it locally in one specific build I add this to the configure line:

cmake ... -DCUDA_USE_STATIC_CUDA_RUNTIME:BOOL=OFF

If I want to set it for all configurations I add this:

option(CUDA_USE_STATIC_CUDA_RUNTIME "use static runtime" OFF)
find_package(CUDA)

Because the first cache variable set will be the winner, you can set it before FindCUDA can set the default.

Modules/FindCUDA.cmake
jamesbigler: I think this would need to have a CMake developer look at it.  I'm not sure adding it to the global search path list is the right way to go about it, and perhaps they can comment on what should be done in your case.  You might already have the best solution.

borisfom: Sure, I'm al for it! Being novice for both CMake and CMake cross-compile, I would definitely appreciate expert advice!

Modules/FindCUDA.cmake
borisfom: Setting this (and a lot of other variables) in top level CMakeLists.txt his is actually option I never tried. Let me try that and see if it works.

borisfom:
I think I should open a new request

PR #221, opened 2016-03-08
Add support for Ubuntu 15.10

olibre:
Other Ubuntu-based distros (and maybe Debian-based ones) should benefit from this commit.
Two other commits for improvement (add path if exist) and cleaning.


bradking:
Thanks!  We prefer to have `Full Name <email@...>` authorship information.  Please rewrite the commits to add it.  Also, please drop the last commit because our coding style generally does not align parens that way.

olibre:
OK I will redo later my commits and I will redo the push request.

PR #222, opened 2016-03-08
Add support for Ubuntu 15.10

olibre:
Other Ubuntu-based distros (and Debian-based ones) should benefit from this commit.
Another commit to consider only existing directories.

This module needs further re-factoring : FindJNI currently searches library and include headers within a list of directories. But one may have installed several java versions and plays with [`update-alternatives`](http://linux.die.net/man/8/update-alternatives) to switch java version. A better way is to retrieve the absolute path of the java command and to deduce the JNI library in the same directory tree... But I do not know how java and JNI are installed on various OS :-/

Something easier I can do for next pull request : Add target `JNI::jni` if you agree... :-)


bradking:
Thanks.  I rebased on `master` and applied these as commit add2557865923af058334ba64a889842fc2abd05 and commit 1b029ea91fe95f3fb036b3b60cf68c2883727ca3.

Adding imported targets would be nice.  See FindPNG for an example.

Please see CONTRIBUTING.rst for the preferred contribution path.

PR #223, opened 2016-03-10
Created modules for Cygwin-Clang

tinysun212:
When building a shared library with clang, the import library was not generated in Cygwin.
The cmake module for GNU can be reused for Clang, because clang accepts almost all of the options for gcc.


bradking:
Thanks.  I revised this as commit 517cef8c5579d271f5af16a385ff22bcfe11f222 using a slightly different approach to re-use the GNU information.  Please verify that it still works for you.

tinysun212:
It works to me. Thanks.

PR #224, opened 2016-03-18
Reworked arch detection file. Added new architectures.

borisfom:


Modules/FindCUDA.cmake
```diff
@@ -586,6 +585,49 @@ if(NOT "${CUDA_SDK_ROOT_DIR}" STREQUAL "${CUDA_SDK_ROOT_DIR_INTERNAL}")
...
+    set(CUDA_TOOLKIT_TARGET_DIR "${CUDA_TOOLKIT_ROOT}/targets/${CUDA_TOOLKIT_TARGET_NAME}" CACHE PATH "CUDA Toolkit target location.")
+    SET (CUDA_TOOLKIT_ROOT_DIR ${CUDA_TOOLKIT_ROOT})
+    mark_as_advanced(CUDA_TOOLKIT_TARGET_DIR)
+  endif()
+
+  # Target CPU architecture: this nvcc option is now obsolete
+  # add known CUDA targetr root path to the set of directories we search for programs, libraries and headers
+  set( CMAKE_FIND_ROOT_PATH "${CUDA_TOOLKIT_TARGET_DIR};${CMAKE_FIND_ROOT_PATH}")
+else()
+ macro( find_host_program )
```
jamesbigler: It's probably a good idea to prefix macros with cuda_, since these will escape the module.

Modules/FindCUDA.cmake
```diff
@@ -647,33 +693,15 @@ else()
   string(REGEX REPLACE "([0-9]+)\\.([0-9]+).*" "\\2" CUDA_VERSION_MINOR "${CUDA_VERSION}")
 endif()
 
+MESSAGE("CUDA_NVCC_EXECUTABLE: ${CUDA_NVCC_EXECUTABLE} CUDA_TOOLKIT_ROOT_DIR: ${CUDA_TOOLKIT_ROOT_DIR} CUDA_VERSION: ${CUDA_VERSION}")
```
jamesbigler: Don't print stuff out.

Modules/FindCUDA.cmake
```diff
@@ -734,7 +764,9 @@ if(NOT CUDA_VERSION VERSION_LESS "5.5")
   cuda_find_library_local_first(CUDA_cudart_static_LIBRARY cudart_static "static CUDA runtime library")
   mark_as_advanced(CUDA_cudart_static_LIBRARY)
 endif()
-if(CUDA_cudart_static_LIBRARY)
+
+
+if(CUDA_USE_STATIC_CUDA_RUNTIME AND CUDA_cudart_static_LIBRARY)
```
jamesbigler: Why are you checking for the value of CUDA_USE_STATIC_CUDA_RUNTIME here?  If someone set this with -D from the commandline or set it via option() before calling this, this check shouldn't be needed.

Modules/FindCUDA.cmake
```diff
@@ -764,15 +796,15 @@ if(CUDA_USE_STATIC_CUDA_RUNTIME)
     else()
       unset(CMAKE_THREAD_PREFER_PTHREAD)
     endif()
-    if (NOT APPLE)
-      # Here is librt that has things such as, clock_gettime, shm_open, and shm_unlink.
-      find_library(CUDA_rt_LIBRARY rt)
-      if (NOT CUDA_rt_LIBRARY)
-        message(WARNING "Expecting to find librt for libcudart_static, but didn't find it.")
-      endif()
+  endif(UNIX AND NOT ANDROID)
```
jamesbigler: Current CMake style is to not put the matching expression in the endif().

Modules/FindCUDA.cmake
```diff
@@ -945,7 +977,9 @@ set(CUDA_TOOLKIT_TARGET_DIR_INTERNAL "${CUDA_TOOLKIT_TARGET_DIR}" CACHE INTERNAL
 set(CUDA_SDK_ROOT_DIR_INTERNAL "${CUDA_SDK_ROOT_DIR}" CACHE INTERNAL
   "This is the value of the last time CUDA_SDK_ROOT_DIR was set successfully." FORCE)
 
-include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
+#include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
+include(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
```
jamesbigler: Use CMAKE_CURRENT_LIST_DIR to get the local copy of this file.  That is the style.  I'm not sure why you changed this.  You also left commented code in here.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+#
+
+# Known NVIDIA GPU achitectures
+# This list will be used for CUDA_ARCH_NAME = All option
+
+set(CUDA_KNOWN_GPU_ARCHITECTURES "2.0 2.1(2.0) 3.0 3.5 5.0")
+
+if (CUDA_VERSION VERSION_GREATER "6.5")
+  set(CUDA_KNOWN_GPU_ARCHITECTURES "3.2 3.7 ${CUDA_KNOWN_GPU_ARCHITECTURES} 5.2")
+  message("CUDA_VERSION VERSION_GREATER 6.5 YES ARCH=${CUDA_KNOWN_GPU_ARCHITECTURES}")
```
jamesbigler: The thing about these messages is that they will print every time you configure.  Most of the time, you shouldn't be printing stuff out each time.

jamesbigler: The "debug" printing should be done by the caller if they are wondering what the CUDA_KNOWN_GPU_ARCHITECTURE are.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+# Variables affecting the choice:
+#
+# CUDA_ARCH_NAME: One of ("Fermi" "Kepler" "Kepler-M" "Maxwell" "Maxwell-M" "All" "Manual")
+# ENV{CUDA_ARCH_BIN} : Only tested if CUDA_ARCH_NAME set to "Manual"
+#
+
+# Known NVIDIA GPU achitectures
+# This list will be used for CUDA_ARCH_NAME = All option
+
+set(CUDA_KNOWN_GPU_ARCHITECTURES "2.0 2.1(2.0) 3.0 3.5 5.0")
```
jamesbigler: Any reason this is a string instead of a list:
set(CUDA_KNOWN_GPU_ARCHITECTURES 2.0 2.1(2.0) 3.0 3.5 5.0)
...
list(APPEND CUDA_KNOWN_GPU_ARCHITECTURES 6.0 6.2)

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+  set(CUDA_KNOWN_GPU_ARCHITECTURES "${CUDA_KNOWN_GPU_ARCHITECTURES} 6.0 6.2")
+  message("CUDA_VERSION VERSION_GREATER 7.5 YES ARCH=${CUDA_KNOWN_GPU_ARCHITECTURES}")
+endif ()
+
+
+################################################################################################
+# A function for automatic detection of GPUs installed  (IF autodetection is enabled)
+# Usage:
+#   CUDA_DETECT_INSTALLED_GPUS(OUT_VARIABLE)
+function(CUDA_DETECT_INSTALLED_GPUS OUT_VARIABLE)
```
jamesbigler: Is it expected that this function will be run every time you configure, or will the user need to guard the call with something to prevent it from being called unnecessarily?  It's just that a system call to compile and run a cuda executable is rather expensive.

jamesbigler: And usually there is something in the cache to prevent it certain operations from running every configure.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+  set(archs_names "Fermi" "Kepler" "Kepler-M" "Tesla" "Maxwell" "Maxwell-M" "Pascal" "All" "Manual")
+  set(archs_name_default "All")
+  if(NOT CMAKE_CROSSCOMPILING)
+    list(APPEND archs_names "Auto")
+    set(archs_name_default "Auto")
+  endif()
+
+  # SET CUDA_ARCH_NAME strings (so it will be seen as dropbox in CMake-Gui)
+  set(CUDA_ARCH_NAME ${archs_name_default} CACHE STRING "Select target NVIDIA GPU achitecture.")
+  SET_property( CACHE CUDA_ARCH_NAME PROPERTY STRINGS "" ${archs_names} )
```
jamesbigler: mixed case is kind of weird and unexpected.  Just make it all lowercase.

jamesbigler: Actually, I'm not sure what you are trying to do with set_property.  What is this meant to do?

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+
+  # verIFy CUDA_ARCH_NAME value
+  list(FIND archs_names ${CUDA_ARCH_NAME} arch_found)
+
+  if(${arch_found} EQUAL -1)
+    message(FATAL_ERROR "Only ${archs_names} architecture names are supported.")
+  endif()
+
+  if(${CUDA_ARCH_NAME} STREQUAL "Manual")
+    set(CUDA_ARCH_BIN ${CUDA_KNOWN_GPU_ARCHITECTURES} CACHE STRING "SpecIFy 'real' GPU architectures to build binaries for, BIN(PTX) format is supported")
```
jamesbigler: typo on SpecIFy.  Looks like something auto upcased this.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+    set(CUDA_ARCH_PTX "50"                     CACHE STRING "SpecIFy 'virtual' PTX architectures to build PTX intermediate code for")
+    mark_as_advanced(CUDA_ARCH_BIN CUDA_ARCH_PTX)
+  else()
+    unSET(CUDA_ARCH_BIN CACHE)
```
jamesbigler: More mixed case.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+  # Allow a user to specify architecture from env
+  if($ENV{CUDA_ARCH_BIN})
+    set(CUDA_ARCH_NAME "Manual")
+    set(CUDA_ARCH_BIN $ENV{CUDA_ARCH_BIN})
+    unSET(CUDA_ARCH_PTX)
+  endif()
+
+  if(${CUDA_ARCH_NAME} STREQUAL "Fermi")
+    set(cuda_arch_bin "2.0 2.1(2.0)")
+  elseif(${CUDA_ARCH_NAME} STREQUAL "Kepler-M")
```
jamesbigler: What is the M modifier mean?

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+
+
+################################################################################################
+# Function for selecting GPU arch flags for nvcc based on CUDA_ARCH_NAME
+# Usage:
+#   SELECT_NVCC_ARCH_FLAGS(out_variable)
+function(CUDA_SELECT_NVCC_ARCH_FLAGS out_variable)
+  # List of arch names
+# CUDA_ARCH_NAME: One of:
+  set(archs_names "Fermi" "Kepler" "Kepler-M" "Tesla" "Maxwell" "Maxwell-M" "Pascal" "All" "Manual")
```
jamesbigler: There should be documentation on these names somewhere.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
+# BorIFrowed from caffe
```
jamesbigler: The main documentation needs to be in FindCUDA where the document generation will parse it out.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,175 @@
...
+# CUDA_ARCH_NAME: One of:
+  set(archs_names "Fermi" "Kepler" "Kepler-M" "Tesla" "Maxwell" "Maxwell-M" "Pascal" "All" "Manual")
+  set(archs_name_default "All")
+  if(NOT CMAKE_CROSSCOMPILING)
+    list(APPEND archs_names "Auto")
+    set(archs_name_default "Auto")
+  endif()
+
+  # SET CUDA_ARCH_NAME strings (so it will be seen as dropbox in CMake-Gui)
+  set(CUDA_ARCH_NAME ${archs_name_default} CACHE STRING "Select target NVIDIA GPU achitecture.")
```
jamesbigler: Note that I tried really hard to make all the "public" variables for CUDA be lists instead of strings.  I think for consistency we should do the same thing.  There are CMake functions for converting lists to strings, so you wouldn't have to change your code here much.

Modules/FindCUDA.cmake
borisfom: find_host_program is a macro defined if CMAKE_CROSSCOMPILING. I am defining it for else() to be used uniformly below.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: Yes, it is guarded inside the function: if(NOT CUDA_GPU_DETECT_OUTPUT) -> cached variable.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: I suspect the original author was making use of those in CMake GUI.So I'd rather leave names here intact.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: "Mobile", for Tegra. It's not an official name, let me look around to check if there is an official one.

Modules/FindCUDA.cmake
jamesbigler: Oh, even more reason to define a cuda version:

if(CMAKE_CROSSCOMPILING)
  # Define our own version of this function for both with and without cross compilation to simplify later calls
  macro(cuda_find_host_program)
    find_host_program( ${ARGN} )
  endmacro()
  ...
else()
  macro( cuda_find_host_program )
    find_program( ${ARGN} )
  endmacro()
  ...
endif()

Modules/FindCUDA/select_compute_arch.cmake
jamesbigler: SET_property is not CMake coding style.  Please change them to set_property.

Modules/FindCUDA/select_compute_arch.cmake
jamesbigler: You should probably add CUDA_GPU_DETECT_OUTPUT to the list of cache variables removed when the version of the toolkit changes in FindCUDA.cmake (see cuda_unset_include_and_libraries).

Modules/FindCUDA.cmake
borisfom: If I only use FindCUDA.cmake module and FindCUDA directory to shadow those of CMake, I am getting this error:
CMake Error at install/share/cmake/torch/FindCUDA.cmake:969 (include):
  include could not find load file:

    /home/bfomitchev/git/torch-android/install/share/cmake/torch/FindPackageHandleStandardArgs.cmake
Call Stack (most recent call first):
  distro/extra/cutorch/CMakeLists.txt:7 (FIND_PACKAGE)
Is there a better way to handle this ?

jamesbigler: OK, but this should be in the CMake distribution where the FindPackageHandleStandardArgs.cmake file should also be.  If you want to test this properly, you need to test it in a CMake installation.  Not in your local copy.

borisfom:
James, any word on the integration?

jamesbigler:
I was waiting for the things I asked for to be fixed.  Are they done and I
just didn't notice?

Modules/FindCUDA/select_compute_arch.cmake
borisfom: done

Modules/FindCUDA/select_compute_arch.cmake
borisfom: fixed

Modules/FindCUDA/select_compute_arch.cmake
borisfom: done.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: Removed.

Modules/FindCUDA.cmake
borisfom: So this one line should be reverted upon integration to CMake.

Modules/FindCUDA.cmake
borisfom: Removed comments throughout

Modules/FindCUDA.cmake
borisfom: I finally did it in a fashion that allows the user to override it with "false". The problem here was that such an override on boolean option is indistinguashable from no override.

Modules/FindCUDA.cmake
borisfom: Done,

Modules/FindCUDA.cmake
borisfom: Done.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: I have changed names to be super-descriptive.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: All mixed cases removed.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: All mixed cases removed.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: All mixed cases removed.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: I think I did it.

Modules/FindCUDA/select_compute_arch.cmake
borisfom: Added link to NVidia document

borisfom:
@jamesbigler : Yes they were all done. I have also added comments for each now.

Modules/FindCUDA.cmake
```diff
@@ -1,4 +1,3 @@
-#.rst:
```
jamesbigler: I think this line is needed by CMake for auto documentation generation stuff.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,168 @@
+# Synopsis:
+# Function for selecting GPU arch flags for nvcc based on CUDA_ARCH_NAME
+# Usage:
+#   SELECT_NVCC_ARCH_FLAGS(out_variable)
```
jamesbigler: I thought I mentioned this in some other place, but the documentation needs to be in FindCUDA.cmake, since that is where the auto-documentation generator will read it.  You should document the new functions and any variables produced by the script.  There are sections for each.

jamesbigler:
Just some documentation updates, but after that I think it is ready.

Modules/FindCUDA.cmake
```diff
@@ -945,7 +969,11 @@ set(CUDA_TOOLKIT_TARGET_DIR_INTERNAL "${CUDA_TOOLKIT_TARGET_DIR}" CACHE INTERNAL
 set(CUDA_SDK_ROOT_DIR_INTERNAL "${CUDA_SDK_ROOT_DIR}" CACHE INTERNAL
   "This is the value of the last time CUDA_SDK_ROOT_DIR was set successfully." FORCE)
 
-include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
+#include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
+#this change is only needed if this file is used standalone. Please revert on PR merge.
```
jamesbigler: I'm not sure they would do the revert on the merge.  I think this needs to be fixed before the pull request.  Either fix it here or in a fork.

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,168 @@
...
+    set(cuda_arch_bin "5.3 5.3")
+  elseif(${CUDA_ARCH_NAME} STREQUAL "Maxwell")
+    set(cuda_arch_bin "5.0 5.2")
+  elseif(${CUDA_ARCH_NAME} STREQUAL "Pascal")
+    set(cuda_arch_bin "6.0 6.2")
+  elseif(${CUDA_ARCH_NAME} STREQUAL "All")
+    set(cuda_arch_bin ${CUDA_KNOWN_GPU_ARCHITECTURES})
+  elseif(${CUDA_ARCH_NAME} STREQUAL "Auto")
+    CUDA_DETECT_INSTALLED_GPUS(cuda_arch_bin)
+  else()  # (${CUDA_ARCH_NAME} STREQUAL "Manual")
```
jamesbigler: It might be useful to verify CUDA_ARCH_NAME == "Manual" or put in a default case with an error message.

```
elseif(${CUDA_ARCH_NAME} STREQUAL "Manual")
  set(cuda_arch_bin ${CUDA_ARCH_BIN})
else()
  message(SEND_ERROR "Unknown CUDA_ARCH_NAME (${CUDA_ARCH_NAME})")
endif()
```

borisfom:
I have opened https://github.com/Kitware/CMake/pull/237 from a new branch for clean integration. Hopefully all concerns are now addressed!

PR #226, opened 2016-03-20
CMakePackageConfigHelpers: allow to specify a relative CMAKE_INSTALL_PREFIX

sergiud:
CMakePackageConfigHelpers fails with an error if `CMAKE_INSTALL_PREFIX` is a relative path. This pull request fixes the problem.

### DIFF
--- a/Modules/CMakePackageConfigHelpers.cmake
+++ b/Modules/CMakePackageConfigHelpers.cmake
@@ -251,6 +251,11 @@ function(CONFIGURE_PACKAGE_CONFIG_FILE _inputFile _outputFile)
     set(absInstallDir "${installPrefix}/${CCF_INSTALL_DESTINATION}")
   endif()
 
+  # Both absInstallDir and installPrefix may still be relative. Make sure both
+  # paths are absolute.
+  get_filename_component (absInstallDir "${absInstallDir}" ABSOLUTE)
+  get_filename_component (installPrefix "${installPrefix}" ABSOLUTE)
+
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  Please try this instead:

```patch
diff --git a/Modules/CMakePackageConfigHelpers.cmake b/Modules/CMakePackageConfigHelpers.cmake
index 206ea7a..d79249a 100644
--- a/Modules/CMakePackageConfigHelpers.cmake
+++ b/Modules/CMakePackageConfigHelpers.cmake
@@ -241,8 +241,10 @@ function(CONFIGURE_PACKAGE_CONFIG_FILE _inputFile _outputFile)
     else()
       message(FATAL_ERROR "INSTALL_PREFIX must be an absolute path")
     endif()
-  else()
+  elseif(IS_ABSOLUTE "${CMAKE_INSTALL_PREFIX")
     set(installPrefix "${CMAKE_INSTALL_PREFIX}")
+  else()
+    get_filename_component(installPrefix "${CMAKE_INSTALL_PREFIX}" ABSOLUTE)
   endif()
 
   if(IS_ABSOLUTE "${CCF_INSTALL_DESTINATION}")
```

sergiud:
Apart from the syntax error in the ``elseif`` line (the closing ``}`` is missing), the patch works for me.

bradking:
Thanks.  Applied as commit fd31feb03f99dad93f3ba14e093480713c1eb3b3.

PR #228, opened 2016-03-22
Better detection of K&R compilers.

ak-ambi:
__CLASSIC_C__ was HP-specific definition. __STDC__ should be more portable.
It should be defined to 1 for compilers conforming to ANSI C and newer standards.

Note: Some old compilers has __STDC__ defined to 0 to inform that they do not support function prototypes.

### DIFF
--- a/Modules/CMakeCCompilerABI.c
+++ b/Modules/CMakeCCompilerABI.c
@@ -2,7 +2,7 @@
 # error "A C++ compiler has been selected for C."
 #endif
 
-#ifdef __CLASSIC_C__
+#if !defined(__STDC__) || __STDC__ == 0
 # define const
 #endif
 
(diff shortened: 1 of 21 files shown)


bradking:
Thanks.  Applied as commit 71f8c91d7188f8d02dbbffd7b1812dc3f1b2accd.

I split the `Source/kwsys` changes out for separate [integration into upstream KWSys](http://review.source.kitware.com/20937).

bradking:
I had to revert the change because MS compilers define `__STDC__` only when a `-Za` option is given.  See documentation [here](https://msdn.microsoft.com/en-us/library/b0084kay.aspx).

If you wish to pursue this issue further, please read `CONTRIBUTING.rst` and come to the mailing list to discuss it.

PR #229, opened 2016-03-23
Added support for IBM XL compiler

erikzenker:
This PR enables the boost module file to find shared libraries compiled by the IBM compiler.


bradking:
Thanks.  Applied as commit 4527b9c41680aab191de7f37eca3fa47f1b07478.

PR #230, opened 2016-03-26
Minor style nitpics

drbo:
Add quotes around CMake variables, add C++ modeline on .h files, and
replace quoted arguments by argument names in the if() command.


bradking:
Thanks for your interest in contributing.  Some of these changes may be useful but would be better discussed with a broader audience on the mailing list.  Please read `CONTRIBUTING.rst` and bring this up for discussion there.

> Add quotes around CMake variables

That may change semantics and so should not be done in a sweeping fashion.  Many of the proposed changes are not necessary because the evaluated variables will not contain any `;` and therefore not trigger argument separation (spaces in the values don't matter).  Some of the changes may be good precautions, but they should be made as individual commits whose commit message explains the reasoning.

> add C++ modeline on .h files

If we are to do this I do not think it should be made part of the copyright notice block.  Options for such marks should be discussed on the mailing list.

> replace quoted arguments by argument names in the if() command.

Similar to the above, these changes should be split into separate commits with explanations.

PR #232, opened 2016-04-06
Use response files to compile obj files with Ninja

jimon:
In our case we have so many include paths that we hit 32kb limit on Windows, so compiling with Ninja fails. I guess there is a same story with make, will investigate make generator later.

Would be nice to get feedback so I can improve contribution if needed.

### DIFF
--- a/Source/cmNinjaTargetGenerator.cxx
+++ b/Source/cmNinjaTargetGenerator.cxx
@@ -585,8 +585,6 @@ cmNinjaTargetGenerator
   this->GetBuildFileStream() << "\n";
 }
 
-extern int calculateCommandLineLengthLimit(int linkRuleLength);
-
 void
 cmNinjaTargetGenerator
 ::WriteObjectBuildStatement(
--- a/Source/cmNinjaTargetGenerator.cxx
+++ b/Source/cmNinjaTargetGenerator.cxx
@@ -340,33 +340,16 @@ cmNinjaTargetGenerator
   vars.ObjectFileDir = "$OBJECT_FILE_DIR";
 
   cmMakefile* mf = this->GetMakefile();
-  
-  const char* forceRspFile = "CMAKE_NINJA_FORCE_RESPONSE_FILE";
-  bool useResponseFile = mf->IsDefinitionSet(forceRspFile) ||
-                         cmSystemTools::GetEnv(forceRspFile) != 0;
 
   std::string flags = "$FLAGS";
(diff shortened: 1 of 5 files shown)


bradking:
For the Makefile generator, grep the source tree for `USE_RESPONSE_FILE_FOR` to see how we conditionally use response files for includes (compiling), object file lists and library lists (linking).

Source/cmNinjaTargetGenerator.cxx
```diff
@@ -340,12 +340,42 @@ ::WriteCompileRule(const std::string& lang)
...
+    cmakeVarLang += lang;
+
+    // build response file name
+    std::string cmakeLinkVar =  cmakeVarLang + "_RESPONSE_FILE_LINK_FLAG";
+    const char * flag = GetMakefile()->GetDefinition(cmakeLinkVar);
+    if(flag) {
+      responseFlag = flag;
+    } else {
+      responseFlag = "@";
+    }
```
bradking: We should be able to just hard-code use of `@` here because the only reason we have `_RESPONSE_FILE_LINK_FLAG` is because we do not know whether to use `@` or `-Wl,@` when invoking a linker.  For the compiler we are always passing the response file directly.

Source/cmNinjaTargetGenerator.cxx
```diff
@@ -340,12 +340,42 @@ ::WriteCompileRule(const std::string& lang)
   vars.ObjectFileDir = "$OBJECT_FILE_DIR";
 
   cmMakefile* mf = this->GetMakefile();
+  
+  const char* forceRspFile = "CMAKE_NINJA_FORCE_RESPONSE_FILE";
+  bool useResponseFile = mf->IsDefinitionSet(forceRspFile) ||
+                         cmSystemTools::GetEnv(forceRspFile) != 0;
```
bradking: Please refactor this lookup out of `WriteLinkStatement` into a dedicated helper method.  We should not need to duplicate the name of the variable or the lookup calls in every call site that needs to know.

jimon:
@bradking, I've fixed code to address your concerns, please have a look.

bradking:
Thanks.  I split the changes into to refactoring commits 24c9106b7b2c3729985b89b06a2732ee3f555022 and f9644a2d1b52fb566bdad9a7211a86a167d73859.  Then I applied the main change as commit 7731e44f87ca9f70335f211a4ef64a1ea68c7f10.

PR #233, opened 2016-04-06
UseJava: use CMAKE_COMMAND instead of direct cmake call

dancing-leaves:


bradking:
Thanks.  Applied as commit fa382ae85ed11cda00fa3163d5547a3ea434883e.

PR #234, opened 2016-04-08
GetPrerequisites: add path of visual studio 10 to 14 in order to find dumpbin.exe.

hevy-CC4477:


bradking:
Thanks!  Applied as commit c79f88b0197eeb927100769ef5c4cef0275acf3a.  I also followed up with commit a1c402910c2010c0ec91d6325b4119ce0fe8379b to try to handle non-default install locations.

PR #236, opened 2016-04-15
Updated findgit to match latest github version

piaoasd123:
github_path "$ENV{LOCALAPPDATA}/Github/PortableGit*/bin" is not longer
valid, and the new path should be set to github_path
"$ENV{LOCALAPPDATA}/Github/PortableGit*/cmd"


bradking:
Thanks.  Applied as commit 19e8752b86a015832b3414f55a298bda9896fc62.

We should keep the old path in case users have older versions installed.  Also the `git.exe` name does not need to be specified explicitly.  CMake will automatically add `.exe` to each name, and we already look for `git`.

PR #237, opened 2016-04-19
Refactored Android(Tegra) support. 

borisfom:
Added automatic compute arch detection module.


Modules/FindCUDA.cmake
```diff
@@ -2,11 +2,10 @@
 # FindCUDA
 # --------
 #
-# Tools for building CUDA C files: libraries and build dependencies.
```
jamesbigler: Why remove this line?

Modules/FindCUDA.cmake
```diff
@@ -2,11 +2,10 @@
...
-#
 # This script locates the NVIDIA CUDA C tools.  It should work on linux,
 # windows, and mac and should be reasonably up to date with CUDA C
 # releases.
+# The approach follows that taken in Caffe: caffe.berkeleyvision.org/
```
jamesbigler: What approach?  And by the way, I'm not sure what you are attributing to Caffe, but perhaps you could be more specific.  Also, it would be nice to have a line break before a new paragraph, since I use autoformatting on comments, and it would merge this with the previous paragraph.  Add an extra '#' line in front of new paragraphs.

Modules/FindCUDA.cmake
```diff
@@ -41,6 +40,17 @@
 #      nvcc in the generated source.  If you compile to PTX and then load the
 #      file yourself, you can mix bit sizes between device and host.
 #
+#   CUDA_ARCH_NAME (Default is "Auto" for regular builds and "Manual" for cross-builds)
```
jamesbigler: It would be helpful to know where this variable is used or what it controls.

Modules/FindCUDA.cmake
```diff
@@ -187,11 +197,8 @@
 #      (e.g. nvcc -Ipath0 -Ipath1 ... ). These paths usually contain other .cu
 #      files.
 #
-#
```
jamesbigler: There was a conscience break between variables and functions.

Modules/FindCUDA.cmake
```diff
@@ -308,8 +322,20 @@
 #   CUDA_nvcuvid_LIBRARY  -- CUDA Video Decoder library.
 #                            Only available for CUDA version 3.2+.
 #                            Windows only.
+#   CUDA_KNOWN_GPU_ARCH_NAMES      -- "Fermi" "Kepler" "Maxwell" "All" "Common" "Manual" "Auto"
```
jamesbigler: Is it possible to follow the format of the other variables' documentation?

These are values provided by the script, and only if you call something specific, right?

Modules/FindCUDA.cmake
borisfom: Restored.

Modules/FindCUDA.cmake
borisfom: Noted. Line removed.

Modules/FindCUDA.cmake
```diff
@@ -41,11 +41,26 @@
 #      nvcc in the generated source.  If you compile to PTX and then load the
 #      file yourself, you can mix bit sizes between device and host.
 #
+#   CUDA_ARCH_NAME (Default is "Auto" for regular builds and "Manual" for cross-builds)
+#   -- Controls the way CUDA_SELECT_NVCC_ARCH_FLAGS function selects CUDA compute architecture flags.
+#      If you are building for local install, "Auto" should work best as it determines local machine compute arch.
```
jamesbigler: These comments need to be wrapped to 80 characters.  This is CMake's requirements (not mine).

Modules/FindCUDA.cmake
jamesbigler: I would prefer this being named something like CUDA_ARCH_TARGETS or CUDA_BUILD_TARGETS or something besides "NAME".  This isn't a name.  This is a variable that drives which architecture targets are produced by CUDA_SELECT_NVCC_ARCH_FLAGS.  I would even prefer it be named something to do with the name of the function that actually cares about it, but perhaps that would be too wordy.

Modules/FindCUDA.cmake
```diff
@@ -41,11 +41,26 @@
...
+#      If you are building for distribution, set to "Common" or "All" to cover common subsets of architectures.
+#      "Manual" makes CUDA_SELECT_NVCC_ARCH_FLAGS to get flags directly from ENV{CUDA_ARCH_BIN}
```
jamesbigler: Do you expect this to read the environment variable every time you configure or just the first time?  It seems messy to rely on the continual presence of this environment variable.

Would it be better to use the environment variable (when defined) to be the default for CUDA_ARCH_NAME?

jamesbigler: You could do something like (note: see my other comments on variable names):

{code}
if(DEFINED $ENV{CUDA_ARCH_BIN})
  set(CUDA_ARCH_NAME $ENV{CUDA_ARCH_BIN} CACHE LIST "documentation")
else
  set(CUDA_ARCH_NAME Auto CACHE LIST "documentation")
endif()
{code}

Modules/FindCUDA.cmake
```diff
@@ -41,11 +41,26 @@
...
+#      Otherwise, list of GPU architecture names expected (semicolon or space separated).
+#      Names: "Fermi;Kepler;Maxwell;Kepler+Tegra;Kepler+Tesla;Maxwell+Tegra;Pascal".
+#      More information: https://en.wikipedia.org/wiki/CUDA
+#
+#   ENV{CUDA_ARCH_BIN} (Only tested if CUDA_ARCH_NAME set to "Manual")
```
jamesbigler: This name needs to be more descriptive.  See comment on CUDA_ARCH_NAME.

borisfom: I have reworked those parameters entirely - environment is preferred way to set them anyway.

Modules/FindCUDA.cmake
borisfom: To me, it sounds messier to change the rules. I think we should always test it and not cache it - otherwise if it's being set from the shell and changed outside, the change won't get through.

Modules/FindCUDA.cmake
borisfom: Done. This restriction is for comments only, right ?

jamesbigler: Right.  And only the documentation section.

Modules/FindCUDA.cmake
jamesbigler: Manual is till a bad name.  It should be something more descriptive about where the value comes from.  It could be ReadFrom_CUDA_ARCH_BIN or ReadFromEnvironment or something like that.  I'm not thrilled by having this controlled by an environment variable.  If the environment changes you wouldn't get any changes in your build until something caused a configure.  The configure could be triggered unintentionally (not mistakenly) and then stuff starts to change.  It is more clear if you want the build configuration to change you run ccmake or cmake again and change a cache (option) variable.  As it is if you want to change the value you have to change the environment variable and then manually run cmake to reconfigure.

borisfom:
Removed enviromnent override.

Modules/FindCUDA.cmake
```diff
@@ -945,7 +983,9 @@ set(CUDA_TOOLKIT_TARGET_DIR_INTERNAL "${CUDA_TOOLKIT_TARGET_DIR}" CACHE INTERNAL
 set(CUDA_SDK_ROOT_DIR_INTERNAL "${CUDA_SDK_ROOT_DIR}" CACHE INTERNAL
   "This is the value of the last time CUDA_SDK_ROOT_DIR was set successfully." FORCE)
 
-include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
+#include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
```
jamesbigler: This got missed somehow.  Please revert these two lines.

jamesbigler:
Aside from the one comment above (which I think was a missed revert), this is good to go.

jamesbigler:
Great work!

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -0,0 +1,151 @@
...
+        list(APPEND cuda_arch_bin ${arch_name})
+      endif()
+    endforeach()
+  endif()
+
+  message(STATUS "Compiling for CUDA architecture(s): ${cuda_arch_bin}")
+
+  # remove dots and convert to lists
+  string(REGEX REPLACE "\\." "" cuda_arch_bin "${cuda_arch_bin}")
+  string(REGEX REPLACE "\\." "" cuda_arch_ptx "${CUDA_ARCH_PTX}")
```
jamesbigler: Now that I look at this a bit more where is CUDA_ARCH_PTX coming from?

borisfom:
Rebased. Should be ready to merge.

jamesbigler:
This looks great!  The only thing I think I would change would be to make the map between Arch names and SMs configurable, but that could certainly be added in the future.  For now, being able to specify things with the SM number is super convenient.

Please ask Kitware to review it and pull it in.

borisfom:
Thanks James!
I have adjusted mappings according to the latest guidelines we worked out at NVidia: last commit.
Kitware: please review an pull!

bradking:
Thanks for your work on this, Boris!  James, thanks for reviewing and guiding Boris to this point.  I will take a look at this as soon as I get a chance.

Modules/FindCUDA.cmake
```diff
@@ -563,8 +581,7 @@ macro(cuda_unset_include_and_libraries)
   unset(CUDA_npps_LIBRARY CACHE)
   unset(CUDA_nvcuvenc_LIBRARY CACHE)
   unset(CUDA_nvcuvid_LIBRARY CACHE)
-
-  unset(CUDA_USE_STATIC_CUDA_RUNTIME CACHE)
+  unset(CUDA_GPU_DETECT_OUTPUT CACHE)
```
bradking: Why is

    unset(CUDA_USE_STATIC_CUDA_RUNTIME CACHE)

removed?

Modules/FindCUDA.cmake
```diff
@@ -750,7 +793,7 @@ else()
 endif()
 
 if(CUDA_USE_STATIC_CUDA_RUNTIME)
-  if(UNIX)
+  if(UNIX AND NOT ANDROID)
```
bradking: CMake does not define an `ANDROID` value.  You could check `CMAKE_SYSTEM_NAME` instead.

bradking:
Thanks.  Applied as commit 7229ae728fa703390b70a9ded5606b01214c14de (with some whitespace changes split out into a parent commit).

PR #241, opened 2016-05-13
Added FP16 detection

borisfom:


Modules/FindCUDA.cmake
```diff
@@ -683,6 +684,24 @@ find_path(CUDA_TOOLKIT_INCLUDE
...
+if (CUDA_VERSION VERSION_GREATER "7.0")
+  set(cuda_fp16_found ${CUDA_TOOLKIT_INCLUDE})
+else()
+  find_path( cuda_fp16_found
+    cuda_fp16.h # Header included in toolkit
+    PATHS ${CUDA_TOOLKIT_INCLUDE}
+    ENV CUDA_PATH
+    ENV CUDA_INC_PATH
+    NO_DEFAULT_PATH
+    )
```
bradking: This `find_path` call will leave behind a cache entry that is publicly visible.  Can `if(EXISTS)` be used here instead?

Modules/FindCUDA.cmake
```diff
@@ -683,6 +684,24 @@ find_path(CUDA_TOOLKIT_INCLUDE
...
+endif()
+
+if (cuda_fp16_found)
+  set(CUDA_HAS_FP16 true CACHE BOOL "CUDA has half-float support")
+else()
+  set(CUDA_HAS_FP16 false CACHE BOOL "CUDA has half-float support")
+endif()
```
bradking: I don't think the `CUDA_HAS_FP16` result variable needs to be cached because it can be re-computed from the other results on every run.  It is not something we expect the user to set, is it?

bradking:
Cc: @jamesbigler

Modules/FindCUDA.cmake
jamesbigler: I agree.  We could use:

if(EXISTS "${CUDA_TOOLKIT_INCLUDE}/cuda_fp16.h") to set the variable.  It doesn't need to be in the cache.

borisfom: Sure let's drop the cache. However, EXISTS was my first try (here: https://github.com/borisfom/CMake/commit/3da716ac0039f1f73cba5ab54d3adf9804d17058) but it would always return false. Was it because of odd operator precedence ?

jamesbigler: I'm not actually familiar with the extra parentheses notation, but I would expect the them to go on the outside:

if(... OR (EXISTS "file"))

But I don't think you need the parentheses at all:

if (CUDA_VERSION VERSION_GREATER "7.0" OR EXISTS "${CUDA_TOOLKIT_INCLUDE}/cuda_fp16.h")

borisfom:
OK the simple test worked after I have removed extra parens; I have quashed in the fix.

jamesbigler:
Looks good to me.

borisfom:
So, looks like we have it ready to go - please merge!

bradking:
Thanks.  Applied as commit cc95001b1cc20367ea127ea09ea91960782e65a3.

PR #243, opened 2016-05-16
update FindGLEW.cmake in order to be able to find glew from a custom directory

hevy-CC4477:


bradking:
There is no need for a separate `GLEW_ROOT` and we generally don't have many `<pkg>_ROOT` variables in find modules.  If you have a custom prefix where things are installed then add it to the [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/v3.5/variable/CMAKE_PREFIX_PATH.html), either in the cache or the environment.  This works for all packages.

PR #247, opened 2016-06-02
Append a library search path found in official Windows prebuilt binaries

zaufi:
So you don't need to specify `BOOST_LIBRARYDIR` along with `BOOST_ROOT` provided.
See also: https://sourceforge.net/projects/boost/files/boost-binaries/

### DIFF
--- a/Modules/FindBoost.cmake
+++ b/Modules/FindBoost.cmake
@@ -209,6 +209,7 @@
 # Copyright 2007      Mike Jackson
 # Copyright 2008      Andreas Pakulat <apaku@gmx.de>
 # Copyright 2008-2012 Philip Lowman <philip@yhbt.com>
+# Copyright 2016 Alex Turbov <i.zaufi@gmail.com>
 #
 # Distributed under the OSI-approved BSD License (the "License");
 # see accompanying file Copyright.txt for details.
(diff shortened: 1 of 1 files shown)


Modules/FindBoost.cmake
```diff
@@ -862,6 +863,37 @@ function(_Boost_MISSING_DEPENDENCIES componentvar extravar)
...
+# so having Boost_ROOT would be enough to specify to find everything.
+#
+macro(_Boost_UPDATE_LIBRARY_SEARCH_DIRS_WITH_PREBUILD_PATHS componentlibvar basedir)
+  if(MSVC)
+    if(CMAKE_CL_64)
+      set(_arch_suffix 64)
+    else()
+      set(_arch_suffix 32)
+    endif()
+    if(MSVC80)
```
bradking: Please check [CMAKE_CXX_COMPILER_ID](https://cmake.org/cmake/help/v3.6/variable/CMAKE_LANG_COMPILER_ID.html), [CMAKE_CXX_COMPILER_VERSION](https://cmake.org/cmake/help/v3.6/variable/CMAKE_LANG_COMPILER_VERSION.html), and [CMAKE_SIZEOF_VOID_P](https://cmake.org/cmake/help/v3.6/variable/CMAKE_SIZEOF_VOID_P.html) instead of the `MSVC*` and `CMAKE_CL_64` variables.  One may be able to compute the versioned directory name from `CMAKE_CXX_COMPILER_VERSION` instead of building a table that needs maintenance.

zaufi:
There is a bunch of references to `MSVC*` in function `_Boost_GUESS_COMPILER_PREFIX`, maybe they should be fixed too?

bradking:
> There is a bunch of references to MSVC* in function _Boost_GUESS_COMPILER_PREFIX, maybe they should be fixed too?

Yes.  That would be a nice cleanup.  If you do that part, please make it a separate preceding commit (still in this PR is fine).  Thanks!

zaufi:
BTW, said 
> compute the versioned directory name from `CMAKE_CXX_COMPILER_VERSION`
did you mean using `math()`?

zaufi:
Heh, I'm not really proficient w/ `git` :( ... now there is few "same" commits...
I'll do a new clone and apply patches tomorrow opening a new ("clean") PR...

bradking:
> did you mean using math()?

Yes.  Note that the IDE version and the `cl` version differ by `6` prior to VS 2015, at which point they now differ by `5`.

> I'll do a new clone and apply patches tomorrow opening a new ("clean") PR...

You can just force-push a branch of the same name to your GitHub fork and this PR will be updated with it.

zaufi:
@bradking, I don't think it worth to have calculations here... First "exception" you've already mention. Here is another one:

    ...
    elseif(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 13.10)
      set(_boost_COMPILER "-vc71")
    elseif(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 13)
      set(_boost_COMPILER "-vc7")
    ...
so it wouldn't be as clear as "naive" way w/ `if`/`elseif` ...

PR #248, opened 2016-06-03
Append a library search path found in official Windows prebuilt binaries (#2)

zaufi:
So you don't need to specify `BOOST_LIBRARYDIR` along with `BOOST_ROOT` provided.
See also: https://sourceforge.net/projects/boost/files/boost-binaries/

### DIFF
--- a/Modules/FindBoost.cmake
+++ b/Modules/FindBoost.cmake
@@ -209,6 +209,7 @@
 # Copyright 2007      Mike Jackson
 # Copyright 2008      Andreas Pakulat <apaku@gmx.de>
 # Copyright 2008-2012 Philip Lowman <philip@yhbt.com>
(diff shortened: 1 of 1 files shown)


bradking:
Thanks. Applied with minor tweaks as commit 3720d5f4a3c6edfdb1533ac60c9ade0adee92e09 and commit c3a281fd327cce529ca3507f00a62608b26627ee.

zaufi:
I wonder, what the reason to test strings like in `bash`?

    if("x${CMAKE_CXX_COMPILER_ID}" STREQUAL "xMSVC")

bradking:
> what the reason to test strings like in bash?

That is to support projects that do not set [CMP0054](https://cmake.org/cmake/help/v3.6/policy/CMP0054.html) to NEW.  There is a variable called `MSVC`, so the need for this is unique to checks against that compiler id.

PR #249, opened 2016-06-04
Added support for user-configurable PBXBuildFile attributes and mig files.

Bekenn:
Updated the Xcode back-end so that the following lines:
```
set_source_files_properties(defs.mig
    PROPERTIES
        XCODE_FILE_ATTRIBUTES "Client;Server"
)
```
...produce the following output:
```
CC05C50A647749C2A8CBD7FA /* /path/to/defs.mig */ = {isa = PBXBuildFile; fileRef = A98241779C4D44C78764B9A3 /* /path-to-defs.mig */; settings = {ATTRIBUTES = (Client,Server); }; };
```

Additionally, the Xcode back-end now provides default support for mig files, leveraging Xcode's built-in behavior.

### DIFF
--- a/Help/manual/cmake-properties.7.rst
+++ b/Help/manual/cmake-properties.7.rst
@@ -344,6 +344,7 @@ Properties on Source Files
    /prop_sf/VS_XAML_TYPE
    /prop_sf/WRAP_EXCLUDE
    /prop_sf/XCODE_EXPLICIT_FILE_TYPE
+   /prop_sf/XCODE_FILE_ATTRIBUTES
    /prop_sf/XCODE_LAST_KNOWN_FILE_TYPE
 
 .. _`Cache Entry Properties`:
--- a/Help/prop_sf/XCODE_FILE_ATTRIBUTES.rst
+++ b/Help/prop_sf/XCODE_FILE_ATTRIBUTES.rst
@@ -1,0 +1,11 @@
+XCODE_FILE_ATTRIBUTES
+---------------------
+
+Add values to the Xcode ``ATTRIBUTES`` setting on its reference to a
+source file.  Among other things, this can be used to set the role on
+a mig file::
+
+  set_source_files_properties(defs.mig
+      PROPERTIES
+          XCODE_FILE_ATTRIBUTES "Client;Server"
+  )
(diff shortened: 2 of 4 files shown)


Bekenn:
I'll be perfectly happy if only the first three commits are accepted; default handling of .mig files may not be desirable since it would be specific to the Xcode back-end.

bradking:
Cc: @gjasny

gjasny:
The first three look sane but would benefit from a test case.

On the fourth I'm not sure if the mf file extension should really map to the MIG language. Brad, what do you think?

gjasny:
I'll take care of the patches by Sunday.

Bekenn:
I'll try and find some time for adding a test case.  Should I drop the fourth commit from the PR?

gjasny:
It does not make a difference because CMake does not use GitHub for PR or Patches.

gjasny:
I polished and pushed the ``xcode-mig-support`` topic branch: https://cmake.org/gitweb?p=stage/cmake.git;a=shortlog;h=refs/heads/xcode-mig-support

The Mach Interface Generator (mig) produces a .h and one or two .c files from one .mig file.

@bradking Is b969b09b9e0be0c884cac811a28e592b3f9ae1c4 a sane thing to do? Or should we revert some of the commits and insist in using ``add_custom_command`` to invoke the mig generator?

Bekenn:
I'm happy without b969b09; I can always explicitly set the LANGUAGE property on the mig file to invoke Xcode's usual handling (if I know that the Xcode generator is being used).  On the other hand, the MSVC generator knows how to handle .rc files, so there is some precedent for this sort of thing.

bradking:
> Is b969b09 a sane thing to do?

No.  Its effects are a partial implementation of `enable_language(MIG)`.

> the MSVC generator knows how to handle .rc files, so there is some precedent for this sort of thing.

Yes, but that is done with an explicit `enable_language(RC)` in `Modules/Platform/Windows-MSVC.cmake` and similar platforms.  The result is that .rc files work not only in the VS IDE generators but also in Makefile and Ninja generators on Windows with the proper toolchains.

Bekenn:
Ah, OK; that makes a lot more sense.

bradking:
After some minor tweaks, these changes have been merged by commit 9e4725560493fdac423102469da6dbe1779383c4.

PR #250, opened 2016-06-07
added OpenProject Button

exmatrikulator:


bradking:
Cc: @clintonstimpson

bradking:
This is a good start.  A few comments:

* This should be enabled only for generators where it makes sense.  For Ninja and Makefile generators there is no IDE to open the project.

* Whatever API is added to `cmGlobalGenerator` to indicate whether the generator supports this could also have a similar method for asking the generator to give us the path to the project file.  That way we don't have to glob for `*.sln`.

* Will this work with Xcode on Mac OS?

exmatrikulator:
1) The button will be disabled, when no project file will be found
2) `cmGlobalGenerator` is only available, if you are configuring your project. If you only select it you can't get any information
3) I have added Xcode

bradking:
> cmGlobalGenerator is only available, if you are configuring your project. If you only select it you can't get any information

Ah, you need to support the use case that we've just selected an already-generated project and want to open it without re-configuring.  That makes sense.

clintonstimpson:
This does not compile for me. 

CMake/Source/QtDialog/CMakeSetupDialog.cxx:516:10: error: no viable conversion from 'bool' to 'QString' 

return false; 

^~~~~

clintonstimpson:
The opening of an Xcode project doesn't work.

It will if I have this code:
void CMakeSetupDialog::doOpenProject()
{
  QString proj = this->getProjectFilename();
  QString open_url = QString("file://%1").arg(proj);
  QDesktopServices::openURL(open_url);
}

Can you verify that with Visual Studio?  I imagine it will work.
After that fix, I believe this patch can be merged.

exmatrikulator:
sorry, it doesn't.

does this work with Xcode?

void CMakeSetupDialog::doOpenProject()
{
	QDesktopServices::openUrl( QUrl::fromLocalFile( this->getProjectFilename() ) );
}

clintonstimpson:
Yes, using QUrl::fromLocalFile() does work with Xcode.

clintonstimpson:
Also, we need a better author name than "ex <ex@exmatrikulator.de>".  I assume we need a first and last name.

I took your 3 commits, included the QUrl fix, and squashed the commits.
When I tried to submit them, I got an error:
$ git push stage HEAD
...
remote: ----------------------------------------------------------------------
remote: commit b3eeb020 has author name "ex" with no space.  Run
remote:   git config --global user.name 'Your Name'
remote:   git config --global user.email 'you@yourdomain.com'
remote: before creating commits.
remote: ----------------------------------------------------------------------
remote: error: hook declined to update refs/heads/open-proj

Without a full name, I assume a Suggested-by: field will work.

bradking:
Applied as commit 1ca2d5d1db5117058e884bcfb226e17e090f29d0.  Thanks1

bradking:
A problem with this was reported on the CMake mailing list [here](https://cmake.org/pipermail/cmake/2016-November/064474.html).

exmatrikulator:
I would suggest to have only one compiler per folder.
Anything else doesn't make sense to me.

PR #251, opened 2016-06-09
Modules: don't assume that FindOpenSSL version regex finds something.

mosra:
Makes the find module work with BoringSSL, which doesn't have the info in the opennslv.h header: https://github.com/google/boringssl/blob/82d0ffbac1a892e23b63d055b67b7d93e489af3e/include/openssl/opensslv.h


bradking:
Thanks.  Applied as commit e937b4c3879e1ee0770b465c0cdcbb6a960ba892.

PR #252, opened 2016-06-09
FindProtobuf: make protobuf_generate_*() work with subdirectories.

mosra:
This now works flawlessly when I have `*.proto` files in subdirectories and pass them to `protobuf_generate_cpp()` like this:
```cmake
protobuf_generate_cpp(App_PB_SRCS App_PB_HEADERS
    services/general.proto
    services/app.proto)
```

Did not encounter any other limitations that would prevent this to work, so I also removed the note from the documentation.


bradking:
Cc: @mathstuf

bradking:
>Did not encounter any other limitations that would prevent this to work

That note is saying that the call to protobuf_generate_cpp and following calls to add_executable/add_library must be in the same directory.  Otherwise the custom commands won't hook up to cause generation to occur.  This has nothing to do with the locations of the generated files.

mosra:
Ah, right, misunderstanding on my side. If I put back the note, the rest is okay?

bradking:
Please squash that in and revise the commit message to explain that it is about `.proto` files in subdirectories.  The change looks okay to me.

mosra:
Done.

Modules/FindProtobuf.cmake
```diff
@@ -197,10 +198,11 @@ function(PROTOBUF_GENERATE_PYTHON SRCS)
   foreach(FIL ${ARGN})
     get_filename_component(ABS_FIL ${FIL} ABSOLUTE)
```
bradking: In the case that there is not a subdirectory path, this may generate double slashes.  Please revise to avoid that.

Modules/FindProtobuf.cmake
```diff
@@ -144,13 +144,14 @@ function(PROTOBUF_GENERATE_CPP SRCS HDRS)
...
     get_filename_component(FIL_WE ${FIL} NAME_WE)
+    get_filename_component(FIL_DIR ${FIL} DIRECTORY)
 
-    list(APPEND ${SRCS} "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.pb.cc")
-    list(APPEND ${HDRS} "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.pb.h")
+    list(APPEND ${SRCS} "${CMAKE_CURRENT_BINARY_DIR}/${FIL_DIR}/${FIL_WE}.pb.cc")
+    list(APPEND ${HDRS} "${CMAKE_CURRENT_BINARY_DIR}/${FIL_DIR}/${FIL_WE}.pb.h")
```
bradking: In the case that there is not a subdirectory path, this may generate double slashes.  Please revise to avoid that.

mosra: Ah, thought that the system would handle this gracefully so I didn't bother.

mosra: Fixed and squashed into d4b8e8135358024b6982100654a430358574efb2.

bradking:
LGTM, thanks.  I've merged to `next` for testing.

PR #253, opened 2016-06-13
Add an option to control what files needs to be traced.

zaufi:
Even in a relatively small projects turning `--trace` (and `--trace-expand`)
may produce a lot of output. But when developing a custom module
usually you are interested in output only one particular module
(ok, few of them, if they interact somehow).

`--trace-file=file` allows to user to specify (multiple options accepted)
what he really wants to trace when `--trace` and/or `--trace-expand` given.
The final output would be only from requested module(s), ignoring anything
else not matched to given filename(s).


bradking:
Cc: @mathstuf

Source/cmake.cxx
```diff
@@ -589,6 +589,10 @@ void cmake::SetArgs(const std::vector<std::string>& args,
       std::cout << "Running with expanded trace output on.\n";
       this->SetTrace(true);
       this->SetTraceExpand(true);
+    } else if (arg.find("--trace-file=", 0) == 0) {
+      std::string file = arg.substr(strlen("--trace-file="));
+      cmSystemTools::ConvertToUnixSlashes(file);
+      this->AddTraceFile(file);
```
mathstuf: Should this call `this->SetTrace(true);` as well? I can't think of a reason it shouldn't.

mathstuf:
Please add docs to `Help/manual/cmake.1.rst`.

Source/cmakemain.cxx
```diff
@@ -73,6 +73,8 @@ static const char* cmDocumentationOptions[][2] = {
   { "--debug-output", "Put cmake in a debug mode." },
   { "--trace", "Put cmake in trace mode." },
   { "--trace-expand", "Put cmake in trace mode with variable expansion." },
+  { "--trace-file=file", "Trace only this CMake file/module. Multiple options "
```
mathstuf: Other argument documentation here uses `<file>` to indicate that it isn't literally `file` (and angle brackets to indicate that it is required).

bradking:
The name `--trace-file=` may also be expected to mean an output file to record the trace.  Let's choose a different name for this option.  Perhaps `--trace-source=`?

Also, this looks like it is intended to match on a file name only.  I don't think that is sufficient because one might want to trace a specific `CMakeLists.txt` file or something.  I think this should work via absolute path.  Relative paths can be converted to full paths during option parsing by preprending the current working directory (and documenting this behavior).  Also use `GetRealPath` on the input paths and the executed paths.  Then use an efficient data structure to perform the lookup during tracing.

Source/cmake.cxx
zaufi: Fixed.

Source/cmakemain.cxx
zaufi: Fixed.

zaufi:
@bradking ,
>  Perhaps --trace-source=?

Fixed.

> I think this should work via absolute path. Relative paths can be converted to full paths during option parsing…

I don't think so… Here is another scenario: you probably don't know the absolute path at the moment of `cmake` execute. Imagine that you have few Foo packages (different versions) with `foo-config.cmake` installed, and you don't really care which would be chosen. Or depending on `CMAKE_MODULE_PATH` content some modules can (or can not) "override" the others (standard). For example until PR #248 not released, I use my "hacked" `FindBoost.cmake` placed in `cmake/` directory in my projects. It is why I've implemented matching  filename only first.

Ok, here is how it can be done:
0) accept any strings (match patterns) as `--trace-source` parameter
1) filename should ends with a given pattern, so at most one may provide an absolute path and it would match. A relative paths, like `libfoo/test/CMakeLists.txt` also would match, if one interested in tracing any particular `CMakeLists.txt`. To prevent useless matching of patterns like `.cmake` or `txt`, `cmSystemTools::GetFilenameName()` of both filename and a pattern should be equal as well.

bradking:
Thanks.  I've applied these changes as commit e63151ff543bb665fea255713e2e643e64527517 and merged to `next` for testing.  Then it will be integrated into `master`.

Source/cmMakefile.cxx
```diff
@@ -181,18 +181,9 @@ void cmMakefile::PrintCommandTrace(const cmListFileFunction& lff) const
...
-        i != trace_only_this_files.end(); ++i) {
-      if (only_filename == *i) {
-        trace = true;
-        break;
-      }
-    }
-  }
+  bool trace = trace_only_this_files.empty() ||
+    std::find(trace_only_this_files.begin(), trace_only_this_files.end(),
+      only_filename) != trace_only_this_files.end();
```
mutanabbi: It's better to us std::any_of (C++11)

mathstuf: FYI, CMake does not use C++11 (without support being hidden behind `#ifdef`'d macros) because it still supports being built compilers which do not have any C++11 support (e.g., Visual Studio 9 2008).

PR #254, opened 2016-06-15
add debug glew to FindGLEW

hevy-CC4477:


bradking:
Thanks.  The proper way to use the debug variants is to find both the release and debug libraries and then select one based on the build configuration.  See the [SelectLibraryConfigurations](https://cmake.org/cmake/help/v3.6/module/SelectLibraryConfigurations.html) module and its uses in some of the other find modules.  See commit b213a7f6ab0d4aa18e7b704bf1cf4994fae77254 for an example.

Also, please read `CONTRIBUTING.rst` for the preferred contribution path.

PR #257, opened 2016-06-22
ninja: Have the bundle resources be a dependency of their target

Orphis:
Bundle resources were added to the "all" target which made it really hard to have OSX bundle resources copied when generating a specific target.
This change adds a dependency on the resources to the target using them.


Orphis:
My guess is that it's too late for 3.6.0 (hence I'm sending to next).
I'm not sure if that's the right way to fix it but it works for me!

Related bug reported 4 years ago: https://cmake.org/Bug/view.php?id=13816

Note that the Xcode and Make generators copy the resources when building the target directly and didn't have that issue (they have other issues though, namely not being Ninja).

Orphis:
@bradking

Orphis:
And after reading the CONTRIBUTING.rst, I realize I should have pushed to master. Will make a new PR then!

PR #258, opened 2016-06-22
ninja: Have the bundle resources be a dependency of their target

Orphis:
Bundle resources were added to the "all" target which made it really hard to have OSX bundle resources copied when generating a specific target.
This change adds a dependency on the resources to the target using them.

### DIFF
--- a/Source/cmNinjaTargetGenerator.cxx
+++ b/Source/cmNinjaTargetGenerator.cxx
@@ -487,6 +487,8 @@ void cmNinjaTargetGenerator::WriteObjectBuildStatements()
   this->GetLocalGenerator()->AppendTargetDepends(this->GeneratorTarget,
                                                  orderOnlyDeps);
 
+  orderOnlyDeps.insert(orderOnlyDeps.end(), ExtraFiles.begin(), ExtraFiles.end());
+
   // Add order-only dependencies on custom command outputs.
   for (std::vector<cmCustomCommand const*>::const_iterator cci =
          this->CustomCommands.begin();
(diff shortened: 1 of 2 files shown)


Orphis:
@bradking 

My guess is that it's too late for 3.6.0 (hence I'm sending to next).
I'm not sure if that's the right way to fix it but it works for me!

Related bug reported 4 years ago: https://cmake.org/Bug/view.php?id=13816

Note that the Xcode and Make generators copy the resources when building the target directly and didn't have that issue (they have other issues though, namely not being Ninja).

bradking:
Thanks for working on this.  For reference, the issue is now tracked [here](https://gitlab.kitware.com/cmake/cmake/issues/13816).

Orphis:
Thanks for pointing to the new bug tracker, Google didn't return results for me when I searched for this issue :)
What do you think of the approach I took? I'm not a big fan of it but it seemed to be the least hacky one.

bradking:
The other generators don't need `cmOSXBundleGenerator` to store anything for them.  The `cmMakefileTargetGenerator` generator builds an `ExtraFiles` list to hook up the dependencies.  I think `cmNinjaTargetGenerator` can do the same.

Orphis:
That is indeed much simpler... I wonder why I didn't do that in the first place.
Amended the commit. Is this better now?

bradking:
Nice.  Applied with minor tweaks as commit 8bbd5db4c160927cf5e9a2eb2f45fe36a2c2d799.  Thanks!

Orphis:
Great! So we'll have this in 3.7.0 in a few months I guess :)

PR #259, opened 2016-06-30
Add missed libraries' check headers.

zaufi:
Official Windows builds of Boost have internal implementation for `zlib` and `bzip2` libraries used by `Boost::iostreams`. Adding headers check would suppress the "_No header defined for…_"
warning.


bradking:
Cc: @rleigh-dundee

rleigh-dundee:
My first question would be exactly how you are using FindBoost to get such an error?  We don't expose the internal embedded zlib or bzip2 at the moment; there is no mention of zlib or bzip2 in the iostreams dependencies or anywhere else.  Are you specifically invoking find_package with zlib or bzip2 as component names?

The current design of FindBoost is that each component is a library.  With your build, do you have boost_zlib and boost_bzip2 DLLs or static libs?  If so, I can see that this change would be useful.

However, I would caution that it also makes your program non-portable since these components will not exist on any other platform.  And they might not even exist on Windows.  (I, for example, build zlib and bzip2, then build boost with these DLLs.  This means my use of FindBoost works across every platform.)

Regarding that portability aim, would it make sense to add support to FindZLib and FindBZip2 to also check for the Boost variant as fallbacks?  If it's drop in compatible with the regular zlib and bzip2 headers and library API/ABI, that would allow you to use find_package with ZLib and BZip2 portably, so that your code will work on Windows with the boost versions, and everywhere else with the regular standalone versions.

That said, the change itself looks completely harmless since it won't be used unless these Windows-specific components are explicitly requested, and so I wouldn't have any reservations about merging this once the question above regarding the intent has been clarified.

Regards,
Roger

zaufi:
> My first question would be exactly how you are using FindBoost to get such an error? We don't expose the internal embedded zlib or bzip2 at the moment; there is no mention of zlib or bzip2 in the iostreams dependencies or anywhere else. Are you specifically invoking find_package with zlib or bzip2 as component names?

Yes, I do. My project uses `iostreams` w/ gzip filters, so my `find_package()` contains it. But, I've faced w/ link errors:
* in \*nix systems I also have to call `find_package(ZLIB)` before `boost` (cuz it uses external implementation) and add it to `target_link_libraries()` among other boost libs
* but in windows, I use `BOOST_ALL_NO_LIB`, so I've add `zlib` as a component to search and use `Boost::zlib` to link with

>The current design of FindBoost is that each component is a library. With your build, do you have boost_zlib and boost_bzip2 DLLs or static libs?

I use [official Windows prebuilt binaries](https://sourceforge.net/projects/boost/files/boost-binaries/1.61.0/). And yes, I have that libs:

    zaufi@gentop〉…/win/boost/lib64-msvc-14.0〉 ls *zip*
    boost_bzip2-vc140-mt-1_59.dll     boost_bzip2-vc140-mt-gd-1_59.lib     libboost_bzip2-vc140-mt-s-1_59.lib    libboost_bzip2-vc140-sgd-1_59.lib
    boost_bzip2-vc140-mt-1_59.lib     libboost_bzip2-vc140-mt-1_59.lib     libboost_bzip2-vc140-mt-sgd-1_59.lib
    boost_bzip2-vc140-mt-gd-1_59.dll  libboost_bzip2-vc140-mt-gd-1_59.lib  libboost_bzip2-vc140-s-1_59.lib

    zaufi@gentop〉…/win/boost/lib64-msvc-14.0〉 ls *zlib?* 
    boost_zlib-vc140-mt-1_59.dll     boost_zlib-vc140-mt-gd-1_59.lib     libboost_zlib-vc140-mt-s-1_59.lib    libboost_zlib-vc140-sgd-1_59.lib
    boost_zlib-vc140-mt-1_59.lib     libboost_zlib-vc140-mt-1_59.lib     libboost_zlib-vc140-mt-sgd-1_59.lib
    boost_zlib-vc140-mt-gd-1_59.dll  libboost_zlib-vc140-mt-gd-1_59.lib  libboost_zlib-vc140-s-1_59.lib

But I use only static libraries of Boost…

> However, I would caution that it also makes your program non-portable since these components will not exist on any other platform. 

Yeah, as I described above, in non Windows, I have to use `FindZLIB`, but Windows build uses `boost_zlib` library available in **official build** -- and thanks to CMake it is easy to define a seamless build process ;)

> I, for example, build zlib and bzip2, then build boost with these DLLs. This means my use of FindBoost/FindZLib/FindBZip2 works across every platform.

Unfortunately I can't afford this way… My goal is to minimize efforts on deploy "build environment" -- it is why I prefer pre-built binaries.

> Regarding that portability aim, would it make sense to add support to FindZLib and FindBZip2 to also check for the Boost variant as fallbacks? 

As I told above, I don't have builds of `ZLIB`/`BZIP2` in Windows… its just another dependency which I have to track/build/deploy to compile cluster and inform other developers that they have to get and install "just another small package". No reason to do that, if Boost already contains all that I need.

>  If it's drop in compatible with the regular zlib and bzip2 headers and library API/ABI, that would allow you to use find_package with ZLib and BZip2 portably, 

In my case it is "cheaper" to use a bit more CMake code to choose depending on platform what to find and what to link with at build time, than having a pure portable way...

zaufi:
@rleigh-dundee, @bradking,

Am I answer all questions of smth else remains? (I have another patch to `FindBoost.cmake` pending for PR and just waiting when this will be closed %)

rleigh-dundee:
@zaufi I have no objections to this being merged.

However, I think there's still a significant missing piece.

Are you using these libraries directly, or are you using them indirectly via boost.iostreams?

For the latter, there's clearly the need for a proper library dependency so that this stuff works automatically by linking with Boost::iostreams.  The complication is that we don't know whether to link with the system zlib/bzip2 or the boost-provided versions (or is there?).  Certainly on Windows, it could be either.  We could certainly add logic to check for the presence of both the system and embedded versions and then add the appropriate dependencies.  We already do this for Boost::threads on Threads::Threads.

Regarding the last question "Regarding that portability aim, would it make sense to add support to FindZLib and FindBZip2 to also check for the Boost variant as fallbacks?", to which you answered "As I told above, I don't have builds of ZLIB/BZIP2 in Windows… its just another dependency which I have to track/build/deploy to compile cluster and inform other developers that they have to get and install "just another small package". No reason to do that, if Boost already contains all that I need." I think you might have misunderstood what I was asking here.  I'd like for this to be completely transparent and work uniformly across all platforms.  I wasn't suggesting that you should install anything.  Merely that the find_package macros should search for both the standalone and boost variants to make this uniform and functional across all platforms, which is what I'm also asking above with regard to exactly how you are using these libraries (direct or indirect via iostreams).  It might be better to do this in the reverse, if the only case to cater for is indirect use via iostreams, by adding the appropriate dependencies to Boost::iostreams.

zaufi:
> Are you using these libraries directly, or are you using them indirectly via boost.iostreams?

The only way for me is to use them directly. First of all, a project I worked on, uses _static_ Boost libraries and have `BOOST_ALL_NO_LIB`.

The other side of the "problem" -- to use or not to use gzip filters of `boost::iostreams` is a configuration option of this project, so depending on it, I have to add (or not) `boost::zlib` to `target_link_libraries()` (and `#define HAVE_GZIP_FEATURE` at compile time) . So, I'm really in doubt how `FindBoost.cmake` can help me here in a "smart way"… In the case I've described, depending on the project option (and thanks to static libraries) I use `boost::iostreams` with or without ZLIB/BZIP2, so how the finder could guess: should it (or shouldn't) find that libs for me… -- it looks too complicated, so simplest (easiest and explicit) solution is to find and add that libs when they really needed.

rleigh-dundee:
Sorry, my question was ambiguous.  Are you using the boost zlib or bzip2 *headers* in your own code or are you using the higher-level iostreams classes?

zaufi:
@rleigh-dundee, the project uses `boost::iostreams` headers only, and `#include` gzip support conditionally.

bradking:
Thanks, folks.  Applied as commit cb1a434ce093ea2530a7acd5692f4da8cfb9f8aa.

PR #263, opened 2016-07-13
regression: fix broken space detection after -D on command line

trivalik:
fix that commandline:
cmake -G "NMake Makefiles" -D CMAKE_C_COMPILER=gcc-4.2

will not use the local CMakeList.txt since the space after -D confuses cmake


bradking:
If this was a regression, in what version did it last work?  Also, please add a case to `Tests/RunCMake/CommandLine/RunCMakeTest.cmake` to cover the behavior.

trivalik:
I guess I made a mistake, because by default you need to specifiy the path where to look for the CMakeList.txt. Somehow I missed that. Sorry.

PR #265, opened 2016-07-18
CPack-STGZ: use correct sub directory

ThomasVolkert:
This fixes directory naming for me.
Best regards.

### DIFF
--- a/Modules/CPack.STGZ_Header.sh.in
+++ b/Modules/CPack.STGZ_Header.sh.in
@@ -98,8 +98,8 @@ ____cpack__here_doc____
   if [ "x${cpack_include_subdir}x" = "xx" ]
   then
     echo "By default the @CPACK_PACKAGE_NAME@ will be installed in:"
-    echo "  \"${toplevel}/@CPACK_PACKAGE_FILE_NAME@\""
-    echo "Do you want to include the subdirectory @CPACK_PACKAGE_FILE_NAME@?"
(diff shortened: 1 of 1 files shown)


bradking:
The convention for binary release tarballs is that the extraction directory matches the name of the tarball.  That is why `CPACK_PACKAGE_FILE_NAME` is used here.  `CPACK_PACKAGE_INSTALL_DIRECTORY` is documented for use by installer tools like NSIS.  The `STGZ` generator is meant to create self-extracting tarballs that are otherwise the same as the `TGZ` generator.

ThomasVolkert:
Okay, but this means that with every new release of the software a new folder name would be used (assuming that CPACK_PACKAGE_FILE_NAME contains some versioning). Is there some way currently available to use the following scheme?:

folder: ${FIXED_NAME}
sh file: ${FIXED_NAME}_${VERSIONING}.sh

bradking:
No, because that breaks the widely-used convention that extracting a tgz produces a directory with the same name.  This convention is not CMake-specific.  It has existed for decades and should be respected.  Among the reasons for this is that extracting different versions in the same working directory should not clobber each other's files.  If one extracts and then manually renames the directory to a version-less name then that latter step will detect such collisions.

If you want a full installer tool then look at the Qt IFW generator.

ThomasVolkert:
Thanks for the feedback. We can skip this patch.

PR #266, opened 2016-07-25
cmake-mode.el: Fix Compilation warning

josteink:
Currently when installing latest cmake-mode in Emacs, you get the following warnings:

````
$ emacs -Q -batch -L . -f byte-compile-file
Byte compile file: cmake-mode.el

In cmake-indent:
cmake-mode.el:141:48:Warning: reference to free variable `cmake-tab-width'
cmake-mode.el:231:1:Warning: defcustom for `cmake-tab-width' fails to specify
    containing group
cmake-mode.el:231:1:Warning: defcustom for `cmake-tab-width' fails to specify
    containing group

In cmake-command-run:
cmake-mode.el:274:8:Warning: `toggle-read-only' is an obsolete function (as of
    24.3); use `read-only-mode' instead.
Wrote d:/Git/CMake/Auxiliary/cmake-mode.elc
````

These are caused by actual errors, incorrect declaration order and use of obsolete functions.

With this patch there are no compilation warnings:

````
$ emacs -Q -batch -L . -f byte-compile-file
Byte compile file: cmake-mode.el
Wrote d:/Git/CMake/Auxiliary/cmake-mode.elc
````

### DIFF
--- a/Auxiliary/cmake-mode.el
+++ b/Auxiliary/cmake-mode.el
@@ -271,7 +271,7 @@ optional argument topic will be appended to the argument list."
     (save-selected-window
       (select-window (display-buffer buffer 'not-this-window))
       (cmake-mode)
-      (toggle-read-only t))
+      (read-only-mode 1))
     )
   )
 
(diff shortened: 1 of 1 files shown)


josteink:
Issued on gitlab.kitware.com instead.

PR #267, opened 2016-08-06
Fixed grep flag for MinGW with LANG != C

GamePad64:
Sometimes, grep assumes, that the output is binary (specifically, when LANG is not C). "-a" flag suppresses this behavior.


bradking:
Thanks.  Applied as commit 5f3c8f6ab2c6633bdbfc083b9e96b477c1700c80.

PR #268, opened 2016-08-12
fix: Uses Matlab's pre-build libraries for MinGW, if needed.

SebastianNiemann:


bradking:
Thanks.  Applied with minor tweak as commit 8ba204a69602ca8744aeec4ca3d808c964715ed5.

PR #269, opened 2016-08-18
Fixing bug in regex not accepting 2.1(2.0) notation

borisfom:
Also, fixed autodetect status message.

### DIFF
--- a/Modules/FindCUDA/select_compute_arch.cmake
+++ b/Modules/FindCUDA/select_compute_arch.cmake
@@ -102,7 +102,7 @@ function(CUDA_SELECT_NVCC_ARCH_FLAGS out_variable)
     set(CUDA_ARCH_LIST ${CUDA_COMMON_GPU_ARCHITECTURES})
   elseif("${CUDA_ARCH_LIST}" STREQUAL "Auto")
     CUDA_DETECT_INSTALLED_GPUS(CUDA_ARCH_LIST)
-    message(STATUS "Autodetected CUDA architecture(s): ${cuda_arch_bin}")
+    message(STATUS "Autodetected CUDA architecture(s): ${CUDA_ARCH_LIST}")
   endif()
 
   # Now process the list and look for names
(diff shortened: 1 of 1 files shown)


borisfom:
@jamesbigler : please take a look!

Modules/FindCUDA/select_compute_arch.cmake
```diff
@@ -116,13 +116,13 @@ function(CUDA_SELECT_NVCC_ARCH_FLAGS out_variable)
...
     endif()
-    if(arch_name MATCHES "([0-9]\\.[0-9])$")
+    if(arch_name MATCHES "(^[0-9]\\.[0-9](\\([0-9]\\.[0-9]\\))?)$")
       set(arch_bin ${CMAKE_MATCH_1})
       set(arch_ptx ${arch_bin})
     else()
       # Look for it in our list of known architectures
       if(${arch_name} STREQUAL "Fermi")
-        set(arch_bin 2.0 "2.1(2.0)")
+        set(arch_bin "2.0 2.1(2.0)")
```
jamesbigler: I thought we decided we wanted arch_bin to be a list.  This makes it into a single string, so when you append it to cuda_arch_bin below, you end up with a list with one element being "2.0 2.1(2.0)".

borisfom: Does it make any difference from here? Resulting flags were just fine. I can revert it.

jamesbigler: If this was unnecessary to change, let's leave it.  I think it makes more sense to try and keep what we have as lists, especially since I see it being used as a list below.

borisfom: Reverted that line.

bradking:
Thanks.  Applied with minor tweak as commit 2dc9a754b539bb215ad2360a27327a629ad26627.

PR #272, opened 2016-09-12
Minor editorial changes

Bekenn:
Escaped backslashes to make them visible in the text.
Renamed VARIABLE to \<var\> in second signature for consistency with the first signature.
Added link to referenced MSDN documentation.


bradking:
Thanks.  Applied with minor tweaks as commit 4f349a8656e4652edacad787d38834255f13a0a5.

PR #273, opened 2016-09-13
[CPack] Generate checksum file with selected algorithm

zerozez:
Often on creating deployment package you also need to write checksum file for it. This PR allows to generate checksum file as soon as package creates without using external tools and stick only with cpack. 

Usage is pretty simple: set **CPACK_PACKAGE_CHECKSUM** by needed algorithm and file will be generated near to package. In case of multiply packages every package will get it own checksum file.

### DIFF
--- a/Modules/CPack.cmake
+++ b/Modules/CPack.cmake
@@ -116,6 +116,15 @@
 #  A branding image that will be displayed inside the installer (used by GUI
 #  installers).
 #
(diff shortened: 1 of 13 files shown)


bradking:
Thanks.  Please open a merge request over on our [GitLab CMake repo](https://gitlab.kitware.com/cmake/cmake) instead.  Your GitHub account can be used to sign in.

PR #274, opened 2016-10-07
FindProtobuf: Fix protobuf_generate_*() to handle subdirs

martjoly:
The change in commit d4b8e81 incorrectly adds subdirectories to the
path of the generated files when *.proto files are passed to
protobuf_generate_*() with subdirectories. This behavior is not correct
when PROTOBUF_GENERATE_CPP_APPEND_PATH is True (default behavior) as
`protoc` will generate output file in the current build directory.


bradking:
Thanks!  I've backported this for merge to the `release` branch for 3.7.0-rc2.  See commit 2b9dad694a43d850af5fe83dd327fc82db29a012.  Once it goes through our nightly testing I'll merge it to `master` and then to `release`.

PR #275, opened 2016-10-10
Addes the default for protobuf_DEBUG_POSTFIX when find debug library

adam4813:
In the call to find_library searching for the debug libraries it doesn't take into account the debug postfix used when building the library using the default provided CMakeLists.txt (example)[https://github.com/google/protobuf/blob/master/cmake/libprotobuf.cmake#L71],

This change adds an alternate name that includes the default postfix of 'd' first in the list of names to find. Since protobuf doesn't use the built-in postfix variable the only choice is to use the default value for the protobuf_DEBUG_POSTFIX variable as that is only set when building protobuf and not when searching for it.

### DIFF
--- a/Modules/FindProtobuf.cmake
+++ b/Modules/FindProtobuf.cmake
@@ -264,7 +264,7 @@ function(_protobuf_find_libraries name filename)
     mark_as_advanced(${name}_LIBRARY_RELEASE)
 
     find_library(${name}_LIBRARY_DEBUG
-      NAMES ${filename}
+      NAMES ${filename}d ${filename}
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  Applied as commit d9190f3308fb67f01f061038a06006af73a77da1.

It looks like their CMake build system provides a `protobuf-config.cmake` file.  This means one can use `find_package(Protobuf CONFIG)` to find it if it is known that the version to be found provides the package configuration file.  That will skip the find module and find the package configuration file directly.  It will also make their imported targets available.  See the [cmake-packages(7)](https://cmake.org/cmake/help/v3.7/manual/cmake-packages.7.html#using-packages) manual for details.

adam4813:
Thanks for the heads up.

PR #278, opened 2016-10-16
add support for CMAKE_SWIG_ODIR to specific output dir for wrap file

congzhangzh:
swig has two output related param:

 -o \<outfile\>    - Set name of C/C++ output file to \<outfile\>
 -outdir \<dir\>   - Set language specific files output directory to \<dir\>

as CMAKE_SWIG_OUTDIR is just for language specific files output directory, we need another for wrap file.

### DIFF
--- a/Modules/UseSWIG.cmake
+++ b/Modules/UseSWIG.cmake
@@ -153,6 +153,14 @@ macro(SWIG_ADD_SOURCE_TO_MODULE name outfiles infile)
   else()
     set(swig_outdir ${CMAKE_CURRENT_BINARY_DIR})
   endif()
+  
+  if(CMAKE_SWIG_ODIR)
+    set(swig_odir ${CMAKE_SWIG_ODIR})
+  else()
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  From the name of the option `CMAKE_SWIG_ODIR` I did not understand its meaning at first.  Also, the use of a `CMAKE_SWIG_` prefix instead of just `SWIG_` is a long-outdated convention.  Let's call the option something like `SWIG_OUTFILE_DIR`.

bradking:
We need to remain compatible with the old option names.  What we can do is honor `SWIG_OUTDIR` if it is set and if not then also check the old `CMAKE_SWIG_OUTDIR` name.  In both cases the documentation should be updated accordingly.

bradking:
@congzhangzh when a revision is ready please open a merge request over in our [GitLab](https://gitlab.kitware.com/cmake/cmake) repository.  That is the preferred contribution path documented in `CONTRIBUTING.rst`.  Your GitHub account can be used to register/authenticate.  Thanks!

congzhangzh:
Ok

PR #280, opened 2016-11-22
Allow FindZLIB to find static libraries made by the zlib project (v1.2.8) cmake build

jeking3:


bradking:
Thanks.  That project has [this line](https://github.com/madler/zlib/blob/v1.2.8/CMakeLists.txt#L204) that makes the library file be just `libz.a` on UNIX platforms, so IIUC this is only for Windows.  Perhaps instead this should just be listed as an alternative name unconditionally.

BTW, please see `CONTRIBUTING.rst` for the preferred contribution path.

PR #283, opened 2017-01-05
Add support for boost 1.63.0.

o01eg:
Add support for new boost version but I'm not sure about dependencies.


sergiud:
It's a bit silly that Boost imported targets and their dependencies must be defined for each Boost release. There are hardly any changes between most releases.

My suggestion would be to always define Boost imported targets regardless of the version and in doubt leave it to the user to specify the dependencies.

purpleKarrot:
There is already a merge request here: https://gitlab.kitware.com/cmake/cmake/merge_requests/361

bradking:
Closing in favor of the GitLab MR 361.

PR #284, opened 2017-01-10
added lib64 to find_library search PATH_SUFFIXES

mtbrobotanist:
Hi, i recently had to add this fix to my system installed FindOpenThreads.cmake module. I had a custom installation of OpenSceneGraph (and OpenThreads) in my home directory. it installed to lib64 and the system supplied FindOpenSceneGraph.cmake was using the system supplied FindOpenThreads.cmake module, which didn't check lib64. after i added lib64, everything compiled fine.

### DIFF
--- a/Modules/FindOpenThreads.cmake
+++ b/Modules/FindOpenThreads.cmake
@@ -86,7 +86,7 @@ find_library(OPENTHREADS_LIBRARY
         /opt/csw
         /opt
         /usr/freeware
-    PATH_SUFFIXES lib
+    PATH_SUFFIXES lib lib64
 )
 
 find_library(OPENTHREADS_LIBRARY_DEBUG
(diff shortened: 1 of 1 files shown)


bradking:
What distro are you using?  The `find_library` command automatically searches `lib64` in place of `lib` on distros that package things that way.  If you're using a custom install please configure it to follow the conventions of the host system (e.g. `lib`) or be prepared to set the library cache entries in your client build with `-DOPENTHREADS_LIBRARY=/path/to/lib64/...`.

We have many find modules that search `lib` directories, and `lib64` is not an appropriate place to search on 32-bit hosts, so hard-coding `lib64` for a specific package is not a good solution.

mtbrobotanist:
Ubuntu 16.04 64bit. OpenSceneGraph installs everything into lib64 by default. Didn't realize that I had that level of control over the installation directory of the libraries specifically.

PR #285, opened 2017-01-18
Xcode schemes

gusc:
I've implemented an Xcode shared scheme writer which mimics the way Xcode generates schemes for all the targets in a project generated by CMake. I've tested it on Xcode 7.0.1, 7.3.1, 8.0, 8.1 and 8.2.1.

This is necessary for automated building and testing using xcodebuild command line instead of Xcode GUI application. The xcodebuild itself does not generate neither private nor shared schemes and thus analyze, archive, build-for-testing and test actions are impossible to run.

I chose to generate shared schemes as they don't require knowing the user's login name. And I chose to create an extra class for schemes because it could be later re-used if for example someone implements Xcode workspace generation (which also can host schemes).

I think that current solution could be advanced further with additional CMake options, but for now it would be enough to just generate a project and be able to build it using command line tools that in some cases require schemes.

If I've forgotten something, please, let me know. :)

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -637,6 +637,7 @@ if(APPLE)
   set(SRCS ${SRCS}
     cmXCodeObject.cxx
     cmXCode21Object.cxx
+    cmXCodeScheme.cxx
     cmGlobalXCodeGenerator.cxx
     cmGlobalXCodeGenerator.h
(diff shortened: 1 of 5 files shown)


gusc:
I'm closing the pull request and moving it to https://gitlab.kitware.com/cmake/cmake

PR #286, opened 2017-02-03
Force generation of folders for Express editions of VS10 and earlier

ahans:
When for compatibility reasons VS10 compilers need to be used, one can still use a newer VS version (supporting folders) as IDE. This patch allows to force the generation of folders even when an Express edition of VS10 is detected by setting the property FORCE_USE_FOLDERS to true.

See also https://cmake.org/pipermail/cmake/2017-January/064942.html


bradking:
I've responded to your message on-list [here](https://cmake.org/pipermail/cmake/2017-February/064971.html).  No force-use-folders feature should be needed.

Please also see `CONTRIBUTING.rst` for the preferred contribution path.

PR #287, opened 2017-02-05
allow override (OPENCL_ROOT) for where comes from

nevion:
We can't expect this script to cover all the different vendors of
OpenCL.  So let's just make a unified variable for users to specify
where we should look.


bradking:
Thanks.  Please see `CONTRIBUTING.rst` for the preferred contribution path and submit a merge request to our GitLab instance.  You can still authenticate with GitHub if preferred.

Also, please add a `Help/release/dev/FindOpenCL-root-var.rst` file with a release note for this feature.  See other notes in that directory for examples.

PR #290, opened 2017-02-16
Add in another name combo for libcrypto that matches libssl.

eyecreate:
libcrypto should have matching find pattern that libssl has for builds without the prefix on the front.


eyecreate:
just read this isn't the right place to submit patches.

PR #291, opened 2017-02-17
Cache the list of xaml and resx headers to speedup VS project generation

devcooch:
Hi!

I've realized that for our huge multi-repository project cmake on Windows runs significantly slower than on Linux.
I've run it in standard VS2015 profiler just to check if there are low-hanging fruits.
Surprisingly GetExpectedResxHeaders and GetExpectedXamlHeaders functions were consuming about 38% of whole CPU time. The reason is that they are triggered for each line when you need to put file inside vcxproj (and check whether it is XAML/RESX header) and they populate a list from scratch.
So I decided to put it in local static var to keep content cached.
For our project the improvement with this change is from 4m44s to 4m12s.
Please let me know if I should use some other approach than static vars.

Regards, 
Dmitry

### DIFF
--- a/Source/cmGeneratorTarget.cxx
+++ b/Source/cmGeneratorTarget.cxx
@@ -711,12 +711,17 @@ void cmGeneratorTarget::GetExternalObjects(
   IMPLEMENT_VISIT(ExternalObjects);
 }
 
-void cmGeneratorTarget::GetExpectedResxHeaders(std::set<std::string>& srcs,
+void cmGeneratorTarget::GetExpectedResxHeaders(std::set<std::string>& headers,
                                                const std::string& config) const
 {
-  ResxData data;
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  Currently we support C++98 host compilers so please avoid using C++11.  Also, please look at caching in a member rather than a static local.

Please see `CONTRIBUTING.rst` for the preferred contribution path and submit this as a MR to our GitLab instance instead.

devcooch:
Clear, thanks! What would you recommend for C++98 compatibility? auto_ptr? Raw pointer deleted in destructor?

bradking:
We use raw pointer deletes in the destructor broadly.  You could try a `CM_AUTO_PTR` member since `cmGeneratorTarget` is a non-copyable type.

devcooch:
The change was merged in GitLab: https://gitlab.kitware.com/cmake/cmake/merge_requests/505

PR #296, opened 2017-03-19
Operation CR - hackers.mu

jmutkawoa:
As per Cryptographic Requirements published on Wikileaks on Marc 2017.
We discard more bytes (3072 bytes) of the first keystream to reduce the possibility of non-random bytes.

### DIFF
--- a/Utilities/cmlibarchive/libarchive/archive_random.c
+++ b/Utilities/cmlibarchive/libarchive/archive_random.c
@@ -222,7 +222,7 @@ arc4_stir(void)
 	 * Discard early keystream, as per recommendations in:
 	 * "(Not So) Random Shuffles of RC4" by Ilya Mironov.
 	 */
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  Please submit this to [upstream libarchive](https://github.com/libarchive/libarchive/) first.  Once reviewed and integrated there we can update CMake's copy.  Otherwise distros that build CMake against their own libarchive won't have this fix.

jmutkawoa:
Hello, its already sent and committed on libarchive/libarchive

jmutkawoa:
@bradking please see https://github.com/libarchive/libarchive/pull/887 for reference

bradking:
Thanks.  I've applied this in our GitLab [MR 598](https://gitlab.kitware.com/cmake/cmake/merge_requests/598).  That is the preferred contribution path described in `CONTRIBUTING.rst`.

PR #299, opened 2017-04-06
FindwxWidgets: link with the new required libs under MSW

vadz:
Latest wxWidgets git master version and the upcoming 3.1.1 release requires
linking with shlwapi and version DLLs. As this does no harm when using the
previous versions, just do it unconditionally.


bradking:
Thanks!

Please see `CONTRIBUTING.rst` for the preferred contribution path and submit a MR to our GitLab instead.  You can still use GitHub OAuth if desired.

vadz:
Sorry for not noticing this, replaced with https://gitlab.kitware.com/cmake/cmake/merge_requests/676

PR #300, opened 2017-04-11
Small fix in 64-bit path for FindVulkan.cmake

Jasper-Bekkers:
The 32-bit path has it (correctly) but the 64-bit version doesn't include NO_SYSTEM_ENVIRONMENT_PATH making it traverse %PATH% and finding an older version of VulkanSDK


Jasper-Bekkers:
Incorrect fix for the problem and wrong place to file it.

PR #303, opened 2017-04-26
Orzuk modify

orzuk-lmj:
We are working on android-nsight with cmake.
This is helpful for us, and lets us to control some options.


bradking:
Thanks.  Please see `CONTRIBUTING.rst` for the preferred contribution path and open a MR on our GitLab instance instead.  You can still use your GitHub account to authenticate via OAuth if desired.

Also, please add documentation for these properties in the `Help/` directory.  Look at examples like `ANDROID_PROCESS_MAX` to follow.

PR #308, opened 2017-06-12
Add -dllimport if has not already been specified

ezralanglois:
SWIG fails with a redefined macro error when `-dllimport` is specified more than once. 

One case where you want to specify your own `-dllimport` is when using SWIG with netstandard2.0. .NET Core does not search for shared libraries like Mono or .NET. So you need to add the shared library prefix and suffix.


bradking:
Thanks!  Please see `CONTRIBUTING.rst` for the preferred contribution path and open a mere request on our GitLab instance instead.  You can use GitHub OAuth to authenticate if preferred.

ezralanglois:
My apologies, I did not read it carefully enough. I will move this over to gitlab.

PR #311, opened 2017-07-25
Fix assertion error of std::string, in testValue

lrineau:
`std::string` cannot be constructed from `nullptr`. With libstdc++, that [throws an assertion][1].

  [1]: https://github.com/vector-of-bool/vscode-cmake-tools/issues/210#issue-242629167


lrineau:
This PR should fix the segfault reported at https://github.com/vector-of-bool/vscode-cmake-tools/issues/210.

bradking:
Thanks.  Ported to [GitLab PR 1096](https://gitlab.kitware.com/cmake/cmake/merge_requests/1096).  Please see `CONTRIBUTING.rst` for the preferred contribution path.

PR #316, opened 2018-01-20
Suggesting changes for source.rst

cpsw:
I would like to suggest to include some additional information on Code Style. This is a new direction towards avoiding certain patterns of code that might increase the changes of adding runtime errors, in my opinion. 

At the same time, I would invite others to share other patterns like these ones with goal of creating a set of guidelines that we could share among open-source projects.

### DIFF
--- a/Help/dev/source.rst
+++ b/Help/dev/source.rst
@@ -47,6 +47,32 @@ building on older toolchains some constructs need to be handled with care:
   the result should be assigned to ``size_t`` not to ``std::size_t``,
   ``unsigned int`` or similar types.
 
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  Please see [CONTRIBUTING.rst](https://gitlab.kitware.com/cmake/cmake/blob/master/CONTRIBUTING.rst) for contribution instructions.  I'll close this in favor of a possible MR over in our GitLab instance.

FWIW, most of the proposed advice is already enforced by lint builds as part of our review and testing infrastructure.

PR #318, opened 2018-01-27
Use as dependency file <objectDir/SourceName>.d if needed.

ClausKlein:
Use as dependency file name simply <objectDir/SourceName>.d if needed
======================================================

Motivation
==========

gcov, lcov
----------

I have a particular problem with the cmake generated output file naming
conventions (like myfile.cpp.o, myfile.cpp.gcno, ....).

These naming conventions creating trouble for me to run `gcov *.cpp` properly
with my -o /objdir option (**I had to copy them in the same directory without
the .ccp extensions every time!**).

see https://cmake.org/pipermail/cmake/2012-August/051628.html

Ancien cross compiler
---------------------

The C compiler test fails because CMake insists on createing
testCCompiler.c.r30  and there doesn't seem to be way to tell it to  not
include the source file name extension (here: .c) into the object  file name.

However, the linker takes the first dot as extension (I know that this  is
horribly broken) and only takes files that have ".r30" as  extension. ".c.r30"
is an illegal extension.

see https://cmake.org/pipermail/cmake/2008-April/021057.html

More proprietary compilers
--------------------------

Ninja generator sets the name to objectpath.d::

    cmGlobalNinjaGenerator::EncodeDepfileSpace(objectFileName + ".d");

WindRiver sets the dependency file name to objectDir/SourceName.d, so the ".obj"
in "DEP_FILE = path/BaseName.c.obj.d" must be renamed!

see https://cmake.org/pipermail/cmake/2014-October/058963.html

GHS compiler too
-----------------

Wenn using the Ninja generator to build with GHS compiler for Integrity targets,
we have the same problem. We can NOT control the name of the generated
dependency ".d" file while compile step!

"DEP_FILE = path/BaseName.d" **has to be used!**

see https://cmake.org/pipermail/cmake-developers/2018-January/030532.html

Solution
========

If we really want to avoid the extension we can set the *undocumented internal
implementation detail* variable::

 set(CMAKE_C_OUTPUT_EXTENSION_REPLACE 1)
 set(CMAKE_CXX_OUTPUT_EXTENSION_REPLACE 1)

some time after the project() command call that enables the C and CXX
languages.

Patch needed
------------

Change the Nina generator::

  file: Source/cmNinjaTargetGenerator.cxx
  function: void cmNinjaTargetGenerator::WriteObjectBuildStatement(...)
  // ...

  if (!this->NeedDepTypeMSVC(language)) {
    bool hasExtension(false);
    std::string dependFileName = this->GetLocalGenerator()->
      GetObjectFileNameWithoutTarget(*source, objectDir, &hasExtension);
    if(hasExtension) {
      // use original code
      vars["DEP_FILE"] = this->GetLocalGenerator()->ConvertToOutputFormat(
        objectFileName + ".d", cmOutputConverter::SHELL);
    } else {
      // Replace the original source file extension with the
      // depend file extension.
      std::string::size_type dot_pos = objectFileName.rfind('.');
      if (dot_pos != std::string::npos) {
        dependFileName = objectFileName.substr(0, dot_pos) + ".d";
      }
      vars["DEP_FILE"] = this->GetLocalGenerator()->ConvertToOutputFormat(
        dependFileName, cmOutputConverter::SHELL);
    }
  }

Caveat
======

As the variables are *internal details* this will not be guaranteed to work in
the future!

Question
========

Is it possible to make the *internal CMAKE_<LANG>_OUTPUT_EXTENSION_REPLACE
option* public and support it in future CMake releases including my patch?

### DIFF
--- a/Source/cmNinjaTargetGenerator.cxx
+++ b/Source/cmNinjaTargetGenerator.cxx
@@ -856,9 +856,25 @@ void cmNinjaTargetGenerator::WriteObjectBuildStatement(
   vars["FLAGS"] = this->ComputeFlagsForObject(source, language);
   vars["DEFINES"] = this->ComputeDefines(source, language);
   vars["INCLUDES"] = this->ComputeIncludes(source, language);
+
   if (!this->NeedDepTypeMSVC(language)) {
-    vars["DEP_FILE"] = this->GetLocalGenerator()->ConvertToOutputFormat(
-      objectFileName + ".d", cmOutputConverter::SHELL);
+    bool hasExtension(false);
+    std::string dependFileName = this->GetLocalGenerator()->
+      GetObjectFileNameWithoutTarget(*source, objectDir, &hasExtension);
+    if(hasExtension) {
+      // use original code
+      vars["DEP_FILE"] = this->GetLocalGenerator()->ConvertToOutputFormat(
+        objectFileName + ".d", cmOutputConverter::SHELL);
+    } else {
+      // Replace the original source file extension with the
+      // depend file extension.
+      std::string::size_type dot_pos = objectFileName.rfind('.');
+      if (dot_pos != std::string::npos) {
+        dependFileName = objectFileName.substr(0, dot_pos) + ".d";
+      }
+      vars["DEP_FILE"] = this->GetLocalGenerator()->ConvertToOutputFormat(
+        dependFileName, cmOutputConverter::SHELL);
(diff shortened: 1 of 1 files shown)


bradking:
Thanks.  Closing in favor of [GitLab MR 1712](https://gitlab.kitware.com/cmake/cmake/merge_requests/1712).  We'll take a look at it over there.

PR #321, opened 2018-06-22
[FindCUDA] add option to blacklist host flags in propagation

SsnL:
Add `CUDA_PROPAGATE_HOST_FLAGS_BLACKLIST`  that specifies a list of flags that shouldn't be propagated when `CUDA_PROPAGATE_HOST_FLAGS` is ON.

Reasoning: propagation is useful, e.g. when specifying debug/release builds. However, sometimes we want some particular flags to not propagated to NVCC, while keeping others. Currently one can only turn `CUDA_PROPAGATE_HOST_FLAGS` ON/OFF, but not filter out specific flags. `CUDA_PROPAGATE_HOST_FLAGS_BLACKLIST` is added to support this.


SsnL:
resubmitting to gitlab

PR #324, opened 2018-11-28
Use a proper way to set FLAGS in UseEcos module

antmak:
I was learning how to set CMAKE_C_FLAGS in different configurations, including cross-compiling toolchains, when I noticed some odd code in the UseEcos module. I'm not 100% sure about change, but maybe you could review this. Thanks. Feel free to throw out it :)


bradking:
Thanks.  Please see [CONTRIBUTING.rst](https://gitlab.kitware.com/cmake/cmake/blob/master/CONTRIBUTING.rst) for contribution instructions and use our gitlab.

Also, please look at using `string(APPEND)`.

PR #327, opened 2019-03-22
Generates Eclipse CDT9 project

dfleury2:
This is an experimental version of a CDT9 project generator based on the CDT4 one.
This version is used for our projects which have particular layout, so it may not works for everybody.
At least CDT9 recognize the imported project as a CDT project and enable the C++ features.
Some improvements can be done like adding all targets...
May be using a C++ template text engine (like Ginger, Pantor, or Amps) could help to generate the XML.


bradking:
Thanks.  Please see [CONTRIBUTING.rst](https://github.com/Kitware/CMake/blob/master/CONTRIBUTING.rst) and open a MR on our GitLab instance.

PR #348, opened 2022-03-15
Update FindCUDAToolkit.cmake

stomov:
Without specifying this to the search path, find_package FindCUDAToolkit can not find the CUDA math libraries (at least this was the case on Perlmutter @ NERSC). I am not sure if
../../math_libs/lib64
should stay in the search path and add this as well:
${CUDAToolkit_TARGET_DIR}/../../math_libs/lib64
or just have the latter. 

Thanks for your interest in contributing to CMake!  The GitHub repository
is a mirror provided for convenience, but CMake does not use GitHub pull
requests for contribution.  Please see

  https://gitlab.kitware.com/cmake/cmake/-/tree/master/CONTRIBUTING.rst

for contribution instructions.  GitHub OAuth may be used to sign in.


bradking:
@stomov the path you updated is meant to be a suffix added to other search paths.  Please open an [issue](https://gitlab.kitware.com/cmake/cmake/-/issues) to report the problem you're encountering for further discussion.

PR #351, opened 2022-05-23
Source: cmFileCommandCurlDebugCallback could modify the return value …

likunyur:
Hello, gitlab doesn't have the permission to submit. The user name is likunyur

### DIFF
--- a/Source/cmFileCommand.cxx
+++ b/Source/cmFileCommand.cxx
@@ -1610,7 +1610,7 @@ size_t cmWriteToMemoryCallback(void* ptr, size_t size, size_t nmemb,
   return realsize;
 }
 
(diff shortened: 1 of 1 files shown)


bradking:
> gitlab doesn't have the permission to submit.

We use the "fork and merge request" model.  You need to us the "Fork" button on the main repo in Gitlab.  Then push your branch to your fork.  Then open a merge request from the fork to the upstream `master` branch.

As for the change itself, [CURLOPT_DEBUGFUNCTION(https://curl.se/libcurl/c/CURLOPT_DEBUGFUNCTION.html) documents that the callback is supposed to have a return value.  However, we've been using `size_t` instead of the documented `int`.

I've opened [CMake MR 7288](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7288) in our GitLab for that.

PR #358, opened 2023-01-23
Improve detection of missing compiler flags

KOLANICH:
move "argument unused during compilation: .*" pattern from target-specific branches into `CMakeCheckCompilerFlagCommonPatterns.cmake`


bradking:
Migrated to [GitLab MR 8117](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8117), thanks.  I also removed one more now-redundant copy of the pattern.

PR #359, opened 2023-02-07
Added additional options to FindwxWidgets

CheesyNacho10:
Added version selection options for better version control of the library. Also added a WIN32 option to select the wxWidgets installation directory by variable.

This can help some wxWidgets users like me that could have more than one wxWidgets installation in their system. Also, for WIN32 users can improve building with CMake because now they don't need to set up an ENV in their systems and can directly use the new custom variable to find the path to their installation.

Hope this makes sense, it did to me :)

### DIFF
--- a/Modules/FindwxWidgets.cmake
+++ b/Modules/FindwxWidgets.cmake
@@ -27,8 +27,11 @@ desired (i.e., these are the only variables you should change to
 select a configuration):
 
 ::
-
-  wxWidgets_ROOT_DIR      - Base wxWidgets directory
+  wxWidgets_CUSTOM_ROOT_DIR
+                          - Base wxWidgets directory
+                            (e.g., C:/ to search C:/wxWidgets-3.2.0).
+  wxWidgets_ROOT_DIR      - Envoirment variable for
+                            base wxWidgets directory
                             (e.g., C:/wxWidgets-3.2.0).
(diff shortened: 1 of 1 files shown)


bradking:
@CheesyNacho10 thanks.  The `find_package` command already defines version request features that FindwxWidgets could be taught to honor.  Please raise your proposed work over in [CMake Issue 10694](https://gitlab.kitware.com/cmake/cmake/-/issues/10694).

Also, FYI, we don't use GitHub PRs for contribution.  See [CONTRIBUTING.rst](https://gitlab.kitware.com/cmake/cmake/blob/master/CONTRIBUTING.rst), which describes how to use our GitLab instance.

CheesyNacho10:
Ok, I'll try to implement that. Where I can find documentation about creating a FindPackage.cmake? I don't know how to extract the argument from the find_package function.

In the other hand, the WIN32 variable for custom folder installation seams reasonable to you?

Whatever are the answer I will provide my solution in the proper site next time :)

bradking:
See the [cmake-developer(7)](https://cmake.org/cmake/help/v3.26/manual/cmake-developer.7.html#find-modules) manual's section on writing find modules.  You could also ask for help over on the [discourse forums](https://discourse.cmake.org/).

CheesyNacho10:
Thanks you so much :D

PR #363, opened 2023-04-05
isUsed standard types

KennethCho-CKH:
Use standard types instead of implementation-defined ones


bradking:
FYI, we don't accept contributions through GitHub MRs.  Please see `CONTRIBUTING.rst`.

However, the change itself is not something we should do.  The `CMakeCCompilerABI.c` source needs to be able to compile with all kinds of compilers, both modern and ancient.  The current code has evolved over many years, and works well.  If you have some case where it fails, please open an issue to describe the problem.

alerque:
This user is playing around with AI generated code, hence the arbitrary nature of theses changes.

PR #372, opened 2024-05-22
Added FindATL.cmake which iss inpired by FindMFC.cmake.

aholzinger:
As MFC and ATL can be installed separately also separate mechanisms to find these are significant.

Sorry, but I don't use GitLab or have a GitLab account.


bradking:
Thanks for the proposal.  However, in general we don't add builtin find modules anymore due to the maintenance burden.

aholzinger:
> Thanks for the proposal. However, in general we don't add builtin find modules anymore due to the maintenance burden.

Well, as there's FindMFC and I needed FindATL, I thought it would be only consequent to have both of these. I was in email contact with Raphael Cazorla and he enncouraged me to create a pull request.

